PR #25, opened 2023-12-07
Added Security Considerations section

cdekater:


draft-dekater-scion-pki.md
```diff
@@ -1303,8 +1335,42 @@ The steps required to create a new AS certificate are the following:
...
+
+Perhaps a more stealthy kill switch would be to shut down control services in victim ASes. An adversarial entity controlling an ISD (e.g., a government) might compel core and non-core ASes to stop replying to path requests. Alternatively, the compelled ASes might return only a subset of all available paths. If this attack were used in conjunction with blackholing, where traffic is redirected to a non-existent resource, senders in the ISD would have difficulty getting traffic out of the ISD. In SCION, however, existing paths can continue to be used in the data plane as long as the traversed ASes allow the forwarding.
+
+
+### Remote ISD/AS Kill Switch
+
+SCION ISDs independently manage their own cryptographic keys and namespace. This prevents a remote attacker who is outside the victim's ISD from causing a kill switch in the victim ISD. Without access to the private keys that form the trust root in the victim ISD, the remote attacker is limited to data-plane attacks. Even if private keys became available to a remote attacker, they would need access to an AS inside the remote ISD to inject faulty information.
```
matzf, line 1359: >  Remote ISD/AS Kill Switch

I think this section / paragraph is redundant. It's already discussed in the introduction of the section that ISDs are isolated.

>Even if private keys became available to a remote attacker, they would need access to an AS inside the remote ISD to inject faulty information.

I don't think that's accurate. There are many ways to bring illegitimately signed AS entries into "circulation".

draft-dekater-scion-pki.md
```diff
@@ -1303,8 +1335,42 @@ The steps required to create a new AS certificate are the following:
...
+
+Another possible attack is when a core AS stops propagating PCBs, thus frustrating the discovery of new paths. In this case, downstream ASes will notice that PCBs are no longer being propagated, but all previously discovered (and still valid) paths are still usable for data-plane forwarding until they expire.
+
+Perhaps a more stealthy kill switch would be to shut down control services in victim ASes. An adversarial entity controlling an ISD (e.g., a government) might compel core and non-core ASes to stop replying to path requests. Alternatively, the compelled ASes might return only a subset of all available paths. If this attack were used in conjunction with blackholing, where traffic is redirected to a non-existent resource, senders in the ISD would have difficulty getting traffic out of the ISD. In SCION, however, existing paths can continue to be used in the data plane as long as the traversed ASes allow the forwarding.
```
matzf, line 1354: Both of these attacks (stop propagating / filter paths) would seem to fit better in the CP section.

draft-dekater-scion-pki.md
```diff
@@ -1303,8 +1335,42 @@ The steps required to create a new AS certificate are the following:
...
+
+DoS attacks, where attackers overload different parts of the IT infrastructure, may impede this process of retrieving missing cryptographic material. SCION offers protection against volumetric DoS attacks, which aim to exhaust network bandwidth on links; in this case, ASes can switch to alternative paths that do not contain the congested links. Transport protocol attacks however, where the attacker tries to exhaust the resources on a target server by opening a large number of connections, may be more difficult to avoid. Possible means to mitigate this kind of DoS attacks are basically the same as for the current Internet, e.g., geo-blocking or using cookies.
```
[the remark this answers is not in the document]
matzf: For certificate renewals, the statement above does not apply (because they are not triggered a CP message).

I think DoS on the certificate renewal infrastructure might need to be discussed explicitly.

draft-dekater-scion-pki.md
```diff
@@ -1303,8 +1335,42 @@ The steps required to create a new AS certificate are the following:
...
+
+
+### Remote ISD/AS Kill Switch
+
+SCION ISDs independently manage their own cryptographic keys and namespace. This prevents a remote attacker who is outside the victim's ISD from causing a kill switch in the victim ISD. Without access to the private keys that form the trust root in the victim ISD, the remote attacker is limited to data-plane attacks. Even if private keys became available to a remote attacker, they would need access to an AS inside the remote ISD to inject faulty information.
+
+
+### Recovery from Kill Switches
+
+In the event of a key compromise of a non-core AS, the impacted AS needs to obtain a new certificate from the core. This process will vary depending on internal issuance protocols. If any of the root keys or voting keys contained in the TRC are compromised, the TRC must be updated as described in [](#update). A trust reset is only required in the case of a catastrophic compromise of multiple voting keys at the same time.
```
matzf, line 1364: I think this would be clearer if we separately discuss compromise of every level of the CP-PKI hierarchy.
The distinction of core/non-core _AS_ is not entirely accurate; the distinction should be whether the AS keys, CA keys, root keys or voting keys have been compromised.

draft-dekater-scion-pki.md
```diff
@@ -1303,8 +1335,42 @@ The steps required to create a new AS certificate are the following:
...
+
+If the core AS has not been compromised, but is instead acting maliciously (e.g., by not propagating PCBs downstream or by falsifying responses to paths- or certificates requests), one way to recover is for downstream ASes to self-organize and form a new ISD. By now operating autonomously, the new ISD can begin path discovery and traffic forwarding. SCION, unlike BGP, has no notion of routing convergence. Instead, the flooding of PCBs disseminates topology information. This means that in the worst case, if all paths must be re-created, fresh paths are established after a single flood has reached all ASes.
```
matzf, line 1366: Again, I think we should focus on the CP-PKI here. We can replace the examples for "acting maliciously" to, "e.g. by not issuing or renewing certificates for certain ASes". 

Reading this though, it's not clear that a core AS acting in such a way can even be considered malicious. It's the job of the core AS to enforce an ISDs policy. So the message is more "If your not happy with an ISDs policy, then don't join this ISD". This seems rather self-evident, to be honest, and doesn't really fit in these security considerations. Maybe we can leave this paragraph away.

draft-dekater-scion-pki.md
```diff
@@ -1303,8 +1335,42 @@ The steps required to create a new AS certificate are the following:
...
+
+
+### Local ISD Kill Switch
+
+Executing a kill switch inside a local ISD can be done at different levels of the AS-level hierarchy. But compared to DNSSEC and BGPsec, a parent authority cannot switch off SCION core ASes, as they manage their own cryptographic trust roots. Another difference with DNSSEC and BGPsec is that the attack vector of intra-ISD kill switches has only two entry levels: All ASes within an ISD obtain certificates directly from the CAs included in the ISD's TRC. If one of the core’s root keys is compromised, an adversary could issue illegitimate AS certificates, which may be used in further attacks. Maliciously change the TRC through a TRC update is more complicated, because each TRC update requires multiple different voting keys (defined by the voting quorum in the TRC).
```
matzf, line 1350: I think this paragraph is the main part of this security considerations section. IMO it would make sense to expand this, discussing the results of compromised keys for every level of the CP-PKI.

cdekater:
Hello @matzf and @jiceatscion,
Thank you very much for your feedback!
Can I now just implement it in the CP-PKI file (although the merge request is still open)?

draft-dekater-scion-pki.md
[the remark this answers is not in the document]
cdekater: @jiceatscion: Thanks for the feedback! I have updated it accordingly.

draft-dekater-scion-pki.md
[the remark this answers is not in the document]
cdekater: @jiceatscion: Thank you for the feedback!
Matthias suggested to remove this paragraph entirely, as it is redundant: It's message is already discussed in the introduction to the Kill Switches section. I think this is a good point.

draft-dekater-scion-pki.md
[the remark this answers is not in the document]
cdekater: @matzf: Thank you very much for your feedback!
This is good point. I have adjusted the draft accordingly (that is, the CP-PKI draft - I still have to revise the security section of the CP draft).

draft-dekater-scion-pki.md
```diff
@@ -1303,8 +1335,42 @@ The steps required to create a new AS certificate are the following:
...
+This section describes the possible security risks and attacks that SCION's control-plane PKI may be prone to, and how these may be mitigated. As SCION is an inter-domain network architecture, the focus lies on *inter*-AS trust issues. All *intra*-AS trust- and security aspects are out of scope.
+
+**Note:** This section only discusses security considerations related to SCION's control-plane PKI. For SCION control plane- and routing-specific security considerations, see {{I-D.scion-cp}}. {{I-D.scion-dp}} includes security considerations that concern the SCION data plane and data forwarding.
+
+
+## Kill Switches
+
+SCION’s trust architecture fundamentally differs from a global monopolistic trust model. In SCION, each ISD manages its own trust roots instead of a single global entity providing those roots. This structure gives each ISD autonomy in terms of key management and in terms of trust. This prevents SCION from the occurrence of a global kill switch affecting all ISDs at once. However, local kill switches are to some extent still possible in SCION. The following sections explain these cases and possible countermeasures.
```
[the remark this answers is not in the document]
cdekater: @matzf: Thank you very much for your feedback and your suggestion!
I have adjusted the draft accordingly.
Best, Corine

draft-dekater-scion-pki.md
cdekater: @matzf: Thank you very much for your feedback!
Good point. I have removed the section from the draft.

Best
Corine

draft-dekater-scion-pki.md
cdekater: @matzf: Thanks again.
Another good point. I have moved both paragraphs to the Security Considerations section of the CP draft.

draft-dekater-scion-pki.md
cdekater: @matzf: Again, good point. I have tried to implement your suggestion, both in this section and in the "Local ISD Kill Switch" section. I am not sure if I got it right, though. It would be great if you could check this. Thanks!
Corine

draft-dekater-scion-pki.md
cdekater: @matzf: good consideration.  I have added a note about it to the text.

draft-dekater-scion-pki.md
cdekater: @matzf: As already mentioned a couple of comments above, I have tried to implement your suggestion about discussing it on each CP-PKI level. As I am not sure if I got it right, it would be great if you could check this. Thanks!

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1344,63 @@ The steps required to create a new AS certificate are the following:
...
+- On AS level: In the event of a key compromise of a (non-core) AS, the impacted AS needs to obtain a new certificate from its CA. This process will vary depending on internal issuance protocols.
+
+If a core AS has not been compromised, but is instead acting maliciously (e.g., by not issuing or renewing certificates for certain ASes), one way to recover is for downstream ASes to self-organize and form a new ISD. By now operating autonomously, the new ISD can begin path discovery and traffic forwarding. SCION, unlike BGP, has no notion of routing convergence. Instead, the flooding of PCBs disseminates topology information. This means that in the worst case, if all paths must be re-created, fresh paths are established after a single flood has reached all ASes.
+
+**Note** that in the above case, it may not be clear whether the core AS acting as described is actually being malicious. It is the job of the core AS to enforce an ISD's policy. If a non-core AS within an ISD does not agree with the in itself legal behavior of the core ASes, it is up to the non-core AS to stay within the ISD - alternatively, the non-core AS could move to another ISD that suits it needs better.
```
knmeynell, line 1378:
```suggestion
**Note** that in the above case, it may not be clear whether the core AS acting as described is actually behaving maliciously. It is the job of the core AS to enforce an ISD's policy. If a non-core AS within an ISD does not agree with ostensibly legal behavior of the core ASes, it is up to the non-core AS to stay within the ISD. Alternatively, the non-core AS could move to another ISD that suits its needs better.
```

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1344,63 @@ The steps required to create a new AS certificate are the following:
...
+
+
+## Denial of Service Attacks
+
+As described previously, the SCION's control-plane PKI lays the foundation for the authentication procedures in SCION. It provides each AS within a specific ISD with a certified key pair. These keys enable the authentication of all control-plane messages - every AS and endpoint can verify all control-plane messages by following the certificate chain.
+
+If an AS realizes that it misses any certificates, or that its certificates are no longer up-to-date (e.g., in case of certificate renewals and revocations), it must query the originator of the message for it. The same goes for an update of a TRC. Updated TRCs are not distributed actively, but must be discovered by the ASes within the ISD. This happens via the processes of beaconing and path-lookup (see also [](#discover-trcupdate)). Upon discovery of a new TRC, an AS must ask the sender of the message including the new TRC number for this new TRC.
```
knmeynell, line 1385:
```suggestion
If an AS realizes that it is missing any certificates, or that its certificates are no longer up-to-date (e.g., in case of certificate renewals and revocations), it must query the originator of the message for these. The same goes for an update of a TRC as updated TRCs are not distributed actively, but must be discovered by the ASes within the ISD. This happens via the processes of beaconing and path-lookup (see also [](#discover-trcupdate)). Upon discovery of a new TRC, an AS must ask the sender of the message including the new TRC number for this new TRC.
```

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1344,63 @@ The steps required to create a new AS certificate are the following:
...
+
+The above implies that the AS can *reach* the originating AS of the message, that is, that paths to the originator are available and that the relevant services at the originating AS are accessible. Denial of Service (DoS) attacks, where attackers overload different parts of the IT infrastructure, may impede this process of retrieving missing cryptographic material. However, the entire process of discovering of and requesting for missing cryptographical material is based on *control-plane* messages and communication. We therefore refer to {{I-D.scion-cp}} for a more detailed description of DoS vulnerabilities of control-plane messages.
```
knmeynell, line 1387:
```suggestion
The above implies that the AS can *reach* the originating AS of the message, that is, that paths to the originator are available and that the relevant services at the originating AS are accessible. Denial of Service (DoS) attacks, where attackers overload different parts of the IT infrastructure, may impede this process of retrieving missing cryptographic material. However, the entire process of discovering and requesting missing cryptographical information is based on *control-plane* messages and communication. We therefore refer to {{I-D.scion-cp}} for a more detailed description of DoS vulnerabilities of control-plane messages.
```

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1344,63 @@ The steps required to create a new AS certificate are the following:
...
 # Security Considerations
 
-The entire document is about security considerations.
-More details will follow in future versions of this draft.
+This section describes the possible security risks and attacks that SCION's control-plane PKI may be prone to, and how these may be mitigated. As SCION is an inter-domain network architecture, the focus lies on *inter*-AS trust issues. All *intra*-AS trust- and security aspects are out of scope.
+
+**Note:** This section only discusses security considerations related to SCION's control-plane PKI. For SCION control plane- and routing-specific security considerations, see {{I-D.scion-cp}}. {{I-D.scion-dp}} includes security considerations that concern the SCION data plane and data forwarding.
+
+
+## Kill Switches
```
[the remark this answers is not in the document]
nicorusti: Right, is "kill-switch" Is this the most appropriate language? Looking at existing RFCs, I see this is seldomly used (but it is used sometimes with IETF. Maybe we can call it "single point of failure"?
Opinions welcome (not a strong one on my side)

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1344,63 @@ The steps required to create a new AS certificate are the following:
...
+
+In a public key infrastructure, the certificate authorities have both the responsibility and power to issue and revoke certificates. If a CA is compromised or malicious, this power can unavoidably be abused. A misbehaving CA can plainly refuse to issue certificates to legitimate entities, or even issue illegitimate certificates to allow impersonating another entity. In the context of the SCION control plane PKI, refusing an AS to obtain or renew their AS certificate will ultimately cut the AS off from the network, turning the CP-PKI into a network "kill switch".
```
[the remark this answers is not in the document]
nicorusti: However there are multiple CAs typically in an ISD, so this is not really a SPOF.  Text needs further adjustment to mention this. 
For example, in productive deployments, there are at least two independent CAs.

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1344,63 @@ The steps required to create a new AS certificate are the following:
...
+
+SCION’s trust architecture fundamentally differs from a global monopolistic trust model. In SCION, each ISD manages its own trust roots instead of a single global entity providing those roots. This structure gives each ISD autonomy in terms of key management and in terms of trust. This prevents SCION from the occurrence of a global kill switch affecting all ISDs at once. However, local ISD kill switches are still possible in SCION. The following sections explain these cases and possible countermeasures.
```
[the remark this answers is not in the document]
nicorusti: We are hinting to RPKI and DNSSEC, but without mentioning it. Maybe we could politely add a mention?

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1344,63 @@ The steps required to create a new AS certificate are the following:
...
+
+
+### Local ISD Kill Switch
+
+Compared to DNSSEC and BGPsec, a parent authority cannot switch off SCION core ASes, as they manage their own cryptographic trust roots. But executing a kill switch *inside* a local ISD can be done at different levels of the AS-level hierarchy:
+
+- On TRC level: The private root keys of the root certificates contained in an TRC are used to sign the CA certificates. If one of these private root keys is compromised, the adversary could issue illegitimate CA certificates, which may be used in further attacks. To maliciously change the entire TRC through a TRC update is more complicated: Each TRC update requires multiple different voting keys. The number of comprimised voting keys needed to perform a malicious update of a TRC depends on the voting quorum set in the TRC. The higher the quorum, the more complicated maliciously updating a TRC will be.
```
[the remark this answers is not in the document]
nicorusti: Right? I'm not fan of the wording here. As far as I understand, compromise of a root certificate means compromise of one of the CAs. Maybe we could reword the paragraph to mention this? 

Also, in principle a CA has  both a root certificate and a CA certificate (according to [here](https://docs.scion.org/en/latest/cryptography/certificates.html#cp-root-certificate), right? Are there cases where a CA certificate is issued to a different organization?

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1344,63 @@ The steps required to create a new AS certificate are the following:
...
+- On CA level: The private keys of an ISD's CA certificates are used to sign the AS certificates. All ASes within an ISD obtain certificates directly from the CAs. If one of the CA’s keys is compromised, an adversary could issue illegitimate AS certificates, which may be used in further attacks.
+- On AS level: Each AS within an ISD signs control-plane messages, such as the Path Construction Beacons (PCBs) used for path discovery, with their AS private key. If the keys of an AS are compromised by an adversary, this adversary can illegitimately sign control-plane messages, including PCBs. This means that the adversary can also manipulate the PCBs, and propagate the manipulated PCBs to neighboring ASes or register/store them as path segments.
+
+
+### Recovery from Kill Switches
```
nicorusti, line 1368:
```suggestion
### Recovery from compromise
```

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1344,63 @@ The steps required to create a new AS certificate are the following:
...
+
+The previous section describes possible "kill switches" on several levels within an ISD. This section deals with possible recovery of these kill switches.
+
+- On TRC level: If any of the root keys or voting keys contained in the TRC are compromised, the TRC must be updated as described in [](#update). Note that this is a sensitive TRC update, as the certificate related to the compromised private key must be replaced with an entirely new certificate (and not just changed). A trust reset is only required in the case of a catastrophic compromise of multiple voting keys at the same time.
+- On CA level: If the private key related to a CA certificate is compromised, the impacted CA AS must obtain a new CA certificate from the corresponding root AS. This process will vary depending on internal issuance protocols.
```
nicorusti, line 1373: Once again, are "root" and CA AS the same? We should clarify this

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1344,63 @@ The steps required to create a new AS certificate are the following:
...
+- On AS level: In the event of a key compromise of a (non-core) AS, the impacted AS needs to obtain a new certificate from its CA. This process will vary depending on internal issuance protocols.
+
+If a core AS has not been compromised, but is instead acting maliciously (e.g., by not issuing or renewing certificates for certain ASes), one way to recover is for downstream ASes to self-organize and form a new ISD. By now operating autonomously, the new ISD can begin path discovery and traffic forwarding. SCION, unlike BGP, has no notion of routing convergence. Instead, the flooding of PCBs disseminates topology information. This means that in the worst case, if all paths must be re-created, fresh paths are established after a single flood has reached all ASes.
```
nicorusti, line 1376: Is there an AS that is not compromised, but acting maliciously?  And is creating another iSD really a solution? I don't think this assumption is realistic and maybe we could remove these lines.

draft-dekater-scion-pki.md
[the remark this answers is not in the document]
nicorusti: Discussed above https://github.com/scionassociation/scion-cppki_I-D/pull/25#discussion_r1422657515

draft-dekater-scion-pki.md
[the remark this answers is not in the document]
nicorusti: I and @knmeynell are also not sure about the term "Kill Switch". It is used sometimes at IETF, but not so much in RFCs. 
Maybe we can use something like single point of failure?

draft-dekater-scion-pki.md
nicorusti: > Reading this though, it's not clear that a core AS acting in such a way can even be considered malicious. 
I agree, maybe we should remove that paragraph.

draft-dekater-scion-pki.md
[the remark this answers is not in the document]
nicorusti: It would be good to mention that the CA certificates are not revoked, but short lived.

draft-dekater-scion-pki.md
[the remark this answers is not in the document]
nicorusti: In the end I moved away from "kill switches" towards key compromise in https://github.com/scionassociation/scion-cppki_I-D/pull/25/commits/6658f2b31a57ad909146eabad82c8b1d59db56ac

draft-dekater-scion-pki.md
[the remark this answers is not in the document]
nicorusti: This is mentioned a bit below. Resolving

draft-dekater-scion-pki.md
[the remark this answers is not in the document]
nicorusti: Added note on multiple CAs in https://github.com/scionassociation/scion-cppki_I-D/pull/25/commits/1a562585fc21a86b0016e22d16f9c8e2c6deafe2

draft-dekater-scion-pki.md
nicorusti: Removed in https://github.com/scionassociation/scion-cppki_I-D/pull/25/commits/6cb1402685fabf6c33241cfb70c065eb203379b2

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1327,57 @@ The steps required to create a new AS certificate are the following:
...
+- On AS level: Each AS within an ISD signs control-plane messages, such as the Path Construction Beacons (PCBs) used for path discovery, with their AS private key. If the keys of an AS are compromised by an adversary, this adversary can illegitimately sign control-plane messages, including PCBs. This means that the adversary can also manipulate the PCBs, and propagate the manipulated PCBs to neighboring ASes or register/store them as path segments.
+
+
+### Recovery from Compromise
+
+This section deals with possible recovery from compromises discussed in the previous paragraph.
+
+- On TRC level: If any of the root keys or voting keys contained in the TRC are compromised, the TRC must be updated as described in [](#update). Note that this is a sensitive TRC update, as the certificate related to the compromised private key must be replaced with an entirely new certificate (and not just changed). A trust reset is only required in the case of a catastrophic compromise of multiple voting keys at the same time.
```
nicorusti, line 1353: @matzf is a trust reset only needed if the number of compromised >= quorum? If so, I'd change the text as follows: 
```suggestion
- On TRC level: If any of the root keys or voting keys contained in the TRC are compromised, the TRC must be updated as described in [](#update). Note that this is a sensitive TRC update, as the certificate related to the compromised private key must be replaced with an entirely new certificate (and not just changed). A trust reset is only required in the case the number of compromised  keys at the same time is greater or equal than the TRC's quorum (see [](#quorum)).
```

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1327,57 @@ The steps required to create a new AS certificate are the following:
...
+
+
+### Compromise of an ISD
+
+Compared to DNSSEC and RPKI, in SCION there is no central authority that could "switch off" an ISD, as each ISD relies on its own independent cryptographic trust roots. Each AS within an ISD is therefore dependant on its ISD's PKI for its functioning. This section discusses potential compromises of the PKI at various levels of the hierarchy.
+
+- On TRC level: The private root keys of the root certificates contained in an TRC are used to sign CA certificates. If one of these private root keys is compromised, the adversary could issue illegitimate CA certificates which may be used in further attacks. To maliciously  perform a TRC update, an attacker would need to compromise multiple voting keys. This number depends on the voting quorum set in the TRC: the higher the quorum, the more unlikely a malicious update will be.
+- On CA level: The private keys of an ISD's CA certificates are used to sign the AS certificates. All ASes within an ISD obtain certificates directly from the CAs. If one of the CA’s keys is compromised, an adversary could issue illegitimate AS certificates, which may be used in further attacks.
```
nicorusti, line 1345: @matzf so there is no revocation for CA certificates?  If one of it is compromised, do you just have to wait that it rolls over (typically 11 days)? Shall we add something more on recovery from compromise?

draft-dekater-scion-pki.md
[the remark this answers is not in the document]
nicorusti: > It would be good to mention that the CA certificates are not revoked, but short lived.


Done in https://github.com/scionassociation/scion-cppki_I-D/pull/25/commits/41bc0a15cb19372761910aa2fa620f7b27cccbdc

draft-dekater-scion-pki.md
matzf: Sounds good to me.

draft-dekater-scion-pki.md
matzf: Correct, there is no revocation, only expiration or TRC updates.
For a CA compromise, the alternative to waiting for expiration is to change the root key by updating the TRC. There is still a large-ish window for attacks during the time it takes for the TRC to be distributed and the grace period. This should be comparably shorter than expiration (I'd assume something on the order of dozens of minutes to few hours at most).

To me, only discussing (the lack of) revocations in the security considerations section does indeed not seem ideal. This is a big design decision and should be discussed front and center.

The book has a short and clearly written section on this topic (3.1.10 Substitutes to Certificate Revocation) that we could adapt. This would fit well into the section  "Trust Relations within an Isolation Domain".

draft-dekater-scion-pki.md
```diff
@@ -1303,21 +1327,57 @@ The steps required to create a new AS certificate are the following:
...
+
+When certificates or TRCs are updated, the relying party must discover and then obtain the new cryptographic material. Thanks to the process of beaconing and path-lookup (see also [](#discover-trcupdate)), relying parties are immediately notified in case a new certificate orn TRCs are available. Upon discovery, the relying party then asks the sender of the message for the new TRC or certificate by reversing the SCION path included in the message received. Reversible paths imply that if a relying party can be notified of  new certificates, then it can also reach the originator of the message to fetch the corresponding cryptographic information.
+
+When it comes to Denial of Service (DoS) attacks, the PKI is therefore dependant on *control-plane* messages and communication. We therefore refer to {{I-D.scion-cp}} for a more detailed description of DoS vulnerabilities of control-plane messages.
```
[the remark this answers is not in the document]
nicorusti: Thank you for rewording! I'm not fully convinced by OOB CA communication. Let's discuss

PR #35, opened 2024-07-07
Clarified use of 'MUST', 'MUST NOT' and 'SHOULD' where required.

nicorusti:
Contributes to https://github.com/scionassociation/scion-cp_I-D/issues/12


draft-dekater-scion-pki.md
```diff
@@ -182,9 +182,9 @@ This document describes the SCION PKI component used by the Control Plane.
...
 
-The monopoly model suffers from two main drawbacks: First, all parties must agree on a single root of trust. Secondly, the single root of trust represents a single point of failure, the misuse of which enables the forging of certificates. Also, its revocation can result in a kill-switch for all the entities it certifies. The oligopoly model relies on several roots of trust, all equally and completely trusted. However, this is not automatically better: Whereas the monopoly model has a single point of failure, the oligopoly model has the drawback of exposing more than one point of failure.
+The monopoly model suffers from two main drawbacks: First, all parties MUST agree on a single root of trust. Secondly, the single root of trust represents a single point of failure, the misuse of which enables the forging of certificates. Also, its revocation can result in a kill-switch for all the entities it certifies. The oligopoly model relies on several roots of trust, all equally and completely trusted. However, this is not automatically better: Whereas the monopoly model has a single point of failure, the oligopoly model has the drawback of exposing more than one point of failure.
```
nicorusti, line 185:
```suggestion
The monopoly model suffers from two main drawbacks: First, all parties must agree on a single root of trust. Secondly, the single root of trust represents a single point of failure, the misuse of which enables the forging of certificates. Also, its revocation can result in a kill-switch for all the entities it certifies. The oligopoly model relies on several roots of trust, all equally and completely trusted. However, this is not automatically better: Whereas the monopoly model has a single point of failure, the oligopoly model has the drawback of exposing more than one point of failure.
```
This is not about SCION, so I'd not use MUST

PR #44, opened 2024-10-08
Deprecate overview draft, add ref to scionlab

knmeynell:


knmeynell:
Which specific wording are we talking about that needs aligning?

nicorusti:
> Which specific wording are we talking about that needs aligning?

The one you took. Thanks!

Issue #56, opened 2025-02-20
TRC discovery: active discovery


nicorusti:
Shall we mention [Active Discovery](https://docs.scion.org/en/latest/cryptography/interactions.html#active-discovery) in `4.1.2.1.   TRC Update Discovery`? Right now we only mention beaconing and path lookups.

@jiceatscion is it actively used?

jiceatscion:
> [@jiceatscion](https://github.com/jiceatscion) is it actively used?

Yes, as far as I know. Fetching TRCs explicitly is part of the API and there are a number of uses of this API in the daemon and control service components.

nicorusti:
Can you add a short paragraph about it?

PR #57, opened 2025-02-26
trc: describe possibilty of explicit TRC query via CS grpc

jiceatscion:
fixes: #56

### DIFF
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -1261,7 +1261,7 @@ In every path segment, all ASes MUST reference the latest TRC of their ISD. Ther
 - *Active Discovery*<br>
 Any TRC can be obtained at any time from the sender of the information it secures; either in a specific version or in its latest available version. The necessary query and response is described in {{I-D.dekater-scion-controlplane}}, section "Control Service gRPC API - Trust Material".
 
-**Note:** The last two mechanisms above only work when there is an active communication between the relying party and the ISD in question.
+**Note:** The last two mechanisms above only work when there is active communication between the relying party and the ISD in question.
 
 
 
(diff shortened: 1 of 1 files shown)


nicorusti (approved):
@jiceatscion do you like my tiny wording changes?

draft-dekater-scion-pki.md
```diff
@@ -1258,7 +1258,10 @@ The TRC version is announced in the beaconing process. Each AS MUST announce wha
 - *Path Lookup*<br>
 In every path segment, all ASes MUST reference the latest TRC of their ISD. Therefore, when resolving paths, every relying party will notice TRC updates, even remote ones.<br>
 
-**Note:** The above mechanism only works when there is an active communication between the relying party and the ISD in question.
+- *Active Discovery*<br>
+Any TRC can be obtained at any time from the sender of the information it secures; either in a specific version or in its latest available version. The necessary query and response is described in {{I-D.dekater-scion-controlplane}}, section "Control Service gRPC API - Trust Material".
+
+**Note:** The last two mechanisms above only work when there is an active communication between the relying party and the ISD in question.
```
knmeynell, line 1264: Propose slight rewording - The last two mechanisms above only work when there is _active_ communication between the relying party and the ISD in question.

nicorusti:
```suggestion
**Note:** The last two mechanisms above only work when there is active communication between the relying party and the ISD in question.
```

jiceatscion: I actually changed the note again: the *FIRST TWO* mechanisms above...
The last one is precisely to cover the lack of active communication.

nicorusti: Ok, therefore the old note was wrong? (it only referred to path lookup)
> **Note:** The above mechanism only works when there is an active communication between the relying party and the ISD in question.

It somehow depends on what you define as active communication

knmeynell: Also, what is a 'specific version' and why would you need it? Wouldn't you always want the latest version, unless there was some sort of broken rollover?

jiceatscion: The API offers both: get a specific version, or get whatever the latest is. I have no idea why a specific version would be desirable. I can imagine some forensic activity; such as checking an old piece of data in its original context?

nicorusti: For the record: if an AS is offline for an extended period of time, it might need to retrieve the whole chain of TRCs. This is why (I think) this is needed

PR #73, opened 2025-07-04
Removed ISD_AS text representation details and refer to CP draft

nicorusti:
Resolves: 
- https://github.com/scionassociation/scion-cp_I-D/issues/112


knmeynell (changes requested):
Suggest

'The string representation of the `ISD-AS number` attribute MUST follow the text representation defined in {{I-D.dekater-scion-controlplane}}, section "Text Representation" where AS numbers in the lower 32-bit range are represented in decimal notation, and others in hexadecimal notation.'

nicorusti:
Applied your suggestion, feel free to merge

PR #88, opened 2025-09-03
Allow other curves

### LINKED ISSUE #76: Post Quantum Cryptography
Ramon Keller (libC) observation:

Only the following three SECP curves are allowed: secp256r1, secp384r1, secp521r1.

The exclusion of other standardized and widely deployed elliptic curves has not been substantiated in the current document. Even if the NIST curves are significantly more efficient, it is conceivable that their origin could be an obstacle for certain SCION networks and that other curves would be preferred.

knmeynell:
Resolves #76


nicorusti (commented):
I think we should figure out some wording about using other curves and interop.

nicorusti (approved):

PR #90, opened 2025-09-03
TRC configuration specification

### LINKED ISSUE #82: Trust Root Configuration Specification
Observation from Ramon Keller (libc):

See "Section 3.1"

The draft just refers to CMS RFC5652 - which is correct - but wouldn't it make sense to directly refer to CMS Signed Data (Section 5. ff)?

knmeynell:
Resolves #82

### DIFF
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -747,7 +747,7 @@ The initial TRC of an ISD is signed during a signing ceremony and then distribut
 
 ## TRC Specification {#trc-spec}
 
-The TRC is a signed collection of {{X.509}} v3 certificates. Additionally, the TRC contains ISD-specific policies encoded in a Cryptographic Message Syntax (CMS) {{RFC5652}} envelope.
+The TRC is a signed collection of {{X.509}} v3 certificates. Additionally, the TRC contains ISD-specific policies encoded in CMS  signed-data ({{RFC5652}} section 5).
 
 The TRC's certificates collection consists of a set of control plane root certificates which build the root of the certification chain for the AS certificates in an ISD. The other certificates in the TRC are solely used for signing the next TRC, a process called "voting". The verification of a new TRC thus depends on the policies and voting certificates defined in the previous TRC.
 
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-pki.md
```diff
@@ -1009,7 +1009,7 @@ That is, the quorum defined in the TRC's `votingQuorum` field ([](#quorum)) MUST
 
 A TRC contains policy information about an ISD and acts as a distribution mechanism for the trust anchors of that ISD.
 
-Each TRC is digitally signed and the syntax used to sign and encapsulate the TRC payload is the Cryptographic Message Syntax (CMS) as defined in {{RFC5652}}. The signed TRC payload is of the CMS signed-data content type, as defined in Section 5 of {{RFC5652}}, and encapsulated in a CMS `ContentInfo` element, as defined in Section 3 of {{RFC5652}}.
+Each TRC is digitally signed and the syntax used to sign and encapsulate the TRC payload is the Cryptographic Message Syntax (CMS) as defined in Section 5 of {{RFC5652}}. The signed TRC payload is encapsulated in a CMS `ContentInfo` element as defined in Section 3 of {{RFC5652}}.
 
 For detailed information on the general syntax definitions of the Cryptographic Message Syntax, see sections 3 and 5 of {{RFC5652}}.
```
nicorusti, line 1015: I think what Ramon suggested (to reference section 5) was already in the text, but later. The problem is that  RFC5652  is referenced too many times. I think we could improve it: 
```suggestion
Each TRC is digitally signed and the syntax used to sign and encapsulate the TRC payload is the Cryptographic Message Syntax (CMS) as defined in Section 5 of {{RFC5652}}. The signed TRC payload is encapsulated in a CMS `ContentInfo` element as defined in Section 3 of {{RFC5652}}.
```

nicorusti: Maybe we could also change the first reference to this RFC: 
```The TRC is a signed collection of [X.509] v3 certificates.
   Additionally, the TRC contains ISD-specific policies encoded in a
   Cryptographic Message Syntax (CMS) [RFC5652] envelope.
   ```

nicorusti (changes requested):

Issue #122, opened 2026-02-06
Final review Nico


nicorusti:
I read through draft-dekater-scion-pk and have a few comments, marked [in here](https://drive.google.com/file/d/10PvbbZ4fe8ejq78fQ5w_VshiRTmHnfa1/view?usp=sharing).

Some TODOs: 

- [ ] #124
- [x] Sections 2 and 3: reduce indents by removing one section
- [x] #128
- [ ] > ~The TRC also provides *trust agility* - that is it enables users to select the trust roots used to initiate certificate validation. This implies that users are free to choose an ISD they believe maintains a uncompromised set of trust roots. . Is this a bit idealistic / theoretical?~
- [x] 4. Deploying the CP PKI - Specifications --> find a better title
- [ ] ~5.2 Initial TRC ceremony: does it belong to 5.2 - Processes?~
- [x] Terminology: check that wording is same as -controlplane / -dataplane
- [ ] ~Voting ASes: do they need to partecipate in routing? (CAs do I think)~
- [x] Table 1: Key chain. Does it add something? Can it be moved to the beginning of section 2 or completely removed?
- [x] Reduce introduction and reference to -controlplane (similar to https://github.com/scionassociation/scion-dp_I-D/pull/133 ) --> #129
- [x] Reduce use of passive tense and be more precise about subjects
- [x] #132
- [x] Deployment Testing: SCIONLab: do we need this section in all drafts? Especially in PKI?

More in the scanned [paper review](url)

nicorusti:
Reopening because there are some comments in https://github.com/scionassociation/scion-cppki_I-D/pull/135 that were not addressed yet. 
I had to merge anyways to have others start with the final pass..

Specifically: 
- [x] Check that @oncilla is ok with wording about authorityCertIssuer & authorityCertSerialNumber in https://github.com/scionassociation/scion-cppki_I-D/pull/135#discussion_r3288789608

PR #123, opened 2026-02-06
Draft review Nico

nicorusti:
Contributes to #122 

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.txt&url_2=https://scionassociation.github.io/scion-cppki_I-D/nic_review/draft-dekater-scion-pki.txt)


draft-dekater-scion-pki.md
```diff
@@ -629,18 +625,13 @@ The `keyUsage` extension identifies the intended usage of the public key in the
...
+
+Other attributes are not used.
 
-**Important:** If a certificate’s public key is used to verify the signature of a control plane payload (`digitalSignature` attribute), it MUST be possible to trace back the private key used to sign the certificate. This is done by referencing the ISD-AS and the subject key identifier (via the `subjectKeyIdentifier` extension). For more information about the `subjectKeyIdentifier` extension, see [](#subject-key-id-ext).
+If a certificate’s public key is used to verify the signature of a control plane payload (`digitalSignature` attribute), it MUST be possible to trace back the private key used to sign the certificate. This is done by referencing the ISD-AS and the subject key identifier (via the `subjectKeyIdentifier` extension). For more information about the `subjectKeyIdentifier` extension (see [](#subject-key-id-ext)).
 
-If present, the `keyUsage` extension SHOULD be marked as "critical". That is, the `critical` Boolean attribute of this extension MUST be set to TRUE (the default is FALSE).
+If present, the `keyUsage` extension's  `critical` attribute MUST be set to TRUE.
```
nicorusti: @oncilla the draft's previous wording was very ambiguous about this (it included a MUST and a SHOULD). I think this is a MUST, correct? 

Or are there any cases where non-critical key usages are expected?

draft-dekater-scion-pki.md
```diff
@@ -704,17 +693,17 @@ The `basicConstraints` extension specifies whether the certificate subject may a
...
 
-| Certificate Type                    | Root                   | CA                            | AS                            | Voting (regular and sensitive)    |
-| ----------------------------------- | ---------------------- | ----------------------------- | ----------------------------- | --------------------------------- |
-| *Attribute:*                        |                        |                               |                               |                                   |
-| `basicConstraints` extension itself | MUST be present        | MUST be present               | SHOULD NOT be present         | SHOULD NOT be present             |
-| `cA`                               | MUST be set to TRUE    | MUST be set to TRUE           | If the extension is present, this attribute MUST be set to FALSE | If the extension is present, this attribute MUST be set to FALSE |
-| `pathLenConstraint`                | SHOULD be set to "1", MUST be marked as "critical" | SHOULD be set to "0" (1), MUST be marked as "critical" | If the extension is present, this attribute MUST be absent. | If the extension is present, this attribute MUST be absent. |
```
nicorusti, line 717: @oncilla is there any reason why this is a SHOULD rather than a MUST? Are there cases when the certificate path is longer than 1?

> `pathLenConstraint`                | SHOULD be set to "1", MUST be marked as "critical" | SHOULD be set to "0"

oncilla: I think we tried to be in line with RFC 5280, but actually in the implementation we have this:

https://github.com/scionproto/scion/blob/68db9a02ff3f7bb0b3cc39217be001cfb2ac6d4f/pkg/scrypto/cppki/certs.go#L305

https://github.com/scionproto/scion/blob/68db9a02ff3f7bb0b3cc39217be001cfb2ac6d4f/pkg/scrypto/cppki/certs.go#L453-L455

So, actually we enforce it AFAICT.

draft-dekater-scion-pki.md
oncilla: Quoting: https://www.rfc-editor.org/rfc/rfc5280.html#section-4.2.1.3

> Conforming CAs MUST include this extension in certificates that
   contain public keys that are used to validate digital signatures on
   other public key certificates or CRLs.  When present, conforming CAs
   SHOULD mark this extension as critical.
   
I think we tried to capture this and be in line with RFC 5280.

nicorusti: Thank you, then this is a SHOULD. 

P.S. the conflict in the old text was: 
> SHOULD be marked as "critical"

Which somehow conflicts with: 
> `critical` Boolean attribute of this extension MUST be set to TRUE

The latest wording is: 
> When present, this extension SHOULD be marked as critical. In such case, the public key in the certificate SHOULD only be used for the purpose set in the critical extension.

draft-dekater-scion-pki.md
nicorusti: So shall I change it to MUST? 
"Must be set to 0" and "MUST be set to 1"

nicorusti:
@knmeynell can you do a thorough 4 eye check (especially on normative language)? Please do not merge yet

draft-dekater-scion-pki.md
```diff
@@ -629,33 +592,26 @@ The `keyUsage` extension identifies the intended usage of the public key in the
...
+Other attributes are not used.
 
-If present, the `keyUsage` extension SHOULD be marked as "critical". That is, the `critical` Boolean attribute of this extension MUST be set to TRUE (the default is FALSE).
+If a certificate’s public key is used to verify the signature of a control plane payload (`digitalSignature` attribute), it MUST be possible to trace back the private key used to sign the certificate. This is done by referencing the ISD-AS and the subject key identifier (via the `subjectKeyIdentifier` extension). For more information about the `subjectKeyIdentifier` extension (see [](#subject-key-id-ext)).
 
-**Note**: If a certificate extension is marked "critical", the public key in the certificate SHOULD only be used for the purpose set in the critical extension.
```
nicorusti, line 645: > **Note**: If a certificate extension is marked "critical", the public key in the certificate SHOULD only be used for the purpose set in the critical extension.

@oncilla I removed this text above because I believe this is duplicating what RFC 5280 already states On KeyUsage:

> If the keyUsage extension is present, then the subject public key
MUST NOT be used to verify signatures on certificates or CRLs unless
the corresponding keyCertSign or cRLSign bit is set.

I think we could therefore piggyback on 5280 and completely remove this sentence on the purpose of the key.

Do you agree?

draft-dekater-scion-pki.md
```diff
@@ -209,13 +176,13 @@ Thus, there is a need for a trust architecture that supports meaningful trust ro
...
 
-Ideally, the trust architecture allows parties that mutually trust each other to form their own trust domain, and to freely decide whether to trust other trust domains outside their own trust bubble.
+Ideally, the trust architecture allows parties that mutually trust each other to form their own trust domain, and to freely decide whether to trust other trust domains.
 
-To fulfill the above requirements, which in fact apply well to inter-domain networking, SCION introduces the concept of **Isolation Domains**. An Isolation Domain (ISD) is a building block for achieving high availability, scalability, and support for heterogeneous trust. It consists of a logical grouping of SCION ASes that share a uniform trust environment (i.e. a common jurisdiction).
+To fulfill the above requirements, which in fact apply well to inter-domain networking, SCION introduces the concept of **Isolation Domains**. An Isolation Domain (ISD) is a building block to support for heterogeneous trust while achieving high availability and scalability in {{I-D.dekater-scion-controlplane}}. It consists of a logical grouping of SCION ASes that share a uniform trust environment (i.e. a common jurisdiction).
```
knmeynell: "to support heterogeneous trust while.."

draft-dekater-scion-pki.md
```diff
@@ -209,13 +176,13 @@ Thus, there is a need for a trust architecture that supports meaningful trust ro
...
+Authentication in SCION is based on X.509 certificates that bind identifiers to public keys and carry digital signatures that are verified by roots of trust. SCION allows each ISD to define its own set of trust roots, along with the policy governing their use. Such scoping of trust roots within an ISD improves security as compromise of a private key associated with a trust root cannot be used to forge a certificate outside the ISD. An ISD's trust roots and policy are encoded in the TRC, which has a version number, a list of public keys that serves as root of trust for various purposes, and a voting quorums governing the number of signatures required to update TRCs. The TRC serves as a way to bootstrap all authentication within SCION. Additionally, TRC versioning is used to efficiently revoke compromised roots of trust.
```
knmeynell: "and a voting quorum..."

draft-dekater-scion-pki.md
```diff
@@ -289,14 +256,14 @@ The base TRC constitutes the root of trust within an ISD. {{figure-1}} provides
...
+All certificates used in the Control Plane PKI are in X.509 v3 format {{RFC5280}} and additionally the TRC contains self-signed certificates instead of plain public keys. Self-signed certificates have the following advantages over plain public keys: (1) They make the binding between name and public key explicit; and (2) the binding is signed to prove possession of the corresponding private key. The public keys of Voting AS certificates must therefore be explicitly verified during the Signing Ceremony ([](#initial-ceremony)).
 
-All ASes in SCION have the task to sign and verify control plane messages. However, certain ASes have additional roles:
+SCION ASes sign and verify control plane messages. Certain ASes have additional roles:
 
-- **Core ASes**: Core ASes are a distinct set of ASes in the SCION Control Plane. For each ISD, the core ASes are listed in the TRC and each core AS has links to the other core ASes (in the same or in different ISDs).
+- **Core ASes**: they are a distinct set of ASes in the SCION Control Plane. For each ISD, the core ASes are listed in the TRC and each core AS has links to the other core ASes (in the same or in different ISDs).
```
knmeynell: Would capitalise 'They'

draft-dekater-scion-pki.md
```diff
@@ -289,14 +256,14 @@ The base TRC constitutes the root of trust within an ISD. {{figure-1}} provides
...
 - **Certification authorities (CAs)**: CAs are responsible for issuing AS certificates to other ASes and/or themselves.
-- **Voting ASes**: Only certain ASes within an ISD may sign TRC updates. The process of appending a signature to a new TRC is called "casting a vote", and the designated ASes that hold the private keys to sign a TRC update are "voting ASes".
-- **Authoritative ASes**: Authoritative ASes are those ASes in an ISD that always have the latest TRCs of the ISD. They start the announcement of a TRC update.
+- **Voting ASes**: they may sign TRC updates. The process of appending a signature to a new TRC is called "casting a vote", and the designated ASes that hold the private keys to sign a TRC update are "voting ASes".
```
knmeynell: Would capitalise 'They'

draft-dekater-scion-pki.md
```diff
@@ -610,17 +573,17 @@ The `authorityKeyIdentifier` extension provides three attributes to specify the
 
 In SCION, using the `keyIdentifier` attribute is the preferred way to specify the `authorityKeyIdentifier` extension.
 
-**Important:** SCION implementations MAY also support the use of the `authorityCertIssuer` and `authorityCertSerialNumber` attributes. However, if these attributes are set and support for them is missing, implementations SHOULD error out.
+SCION implementations MAY also support the use of the `authorityCertIssuer` and `authorityCertSerialNumber` attributes. However, if these attributes are set and support for them is missing, implementations SHOULD error out.
 
-This extension MUST always be non-critical. However, SCION implementations MUST error out if the extension is not present AND the certificate is not self-signed.
+This extension MUST be marked as non-critical. Implementations MUST error out if the extension is not present AND the certificate is not self-signed.
```
knmeynell: I'm not sure 'error out' is good terminology here. Suggest 'Implementations must generate/create an error if the extension is not present...'

draft-dekater-scion-pki.md
```diff
@@ -610,17 +573,17 @@ The `authorityKeyIdentifier` extension provides three attributes to specify the
...
 
 #### `subjectKeyIdentifier` Extension {#subject-key-id-ext}
 
 The `subjectKeyIdentifier` extension identifies certificates that contain a particular public key. It can be used, for example, by control plane messages to identify which certificate to use for verification. The extension allows for overlapping control plane CA keys, for example during updates.
 
 For the syntax and definition of the `subjectKeyIdentifier` extension, see {{RFC5280}}, section 4.2.1.2, and {{X.509}}, clause 9.2.2.2.
 
-This extension MUST always be non-critical. However, SCION implementations MUST error out if the extension is not present.
+This extension MUST be marked as non-critical. Implementations MUST error out if the extension is not present.
```
knmeynell: Same comment about 'error out'.

draft-dekater-scion-pki.md
```diff
@@ -1103,39 +1054,39 @@ The selection algorithm for building the trust anchor pool is described in pseud
...
+All non-base TRCs of an ISD are updates of the ISD's base TRC(s) and constitute a chain. Updates are categorized as regular or sensitive, depending on which payload fields are being modified.
 
-This section describes the rules that apply to updating a TRC in regard to the payload information contained in the TRC. Some rules are valid for both update types whilst some only apply to a regular or a sensitive TRC update. Based on the type of update, different sets of voters are needed to create a verifiable TRC update and the corresponding voting (signing) process is also described. To verify a TRC update, a relying party MUST perform a couple of checks which are also listed.
+This section describes the rules that apply to updating a TRC in regard to the payload information contained in the TRC. Some rules are valid for both update types whilst some only apply to a regular or a sensitive TRC update. Based on the type of update, different sets of voters are needed to create a verifiable TRC update and the corresponding voting (signing) process is also described. Last, this section describes checks to verify a newly issued TRC.
```
knmeynell: I suggest 'Finally' before 'this section describes check.."

draft-dekater-scion-pki.md
```diff
@@ -1179,52 +1130,46 @@ Furthermore, if one or more *new* certificates are added to the updated TRC, the
...
 To verify a TRC update, the relying party MUST perform the following checks:
 
 - Check that the specified update rules as described above are respected.
-- Check whether the update is regular or sensitive.
-   - In case of a regular update,
-      - check that the signatures for the changing certificates are present and verifiable, and
-      - check that all votes are cast by a regular voting certificate.
-   - In case of a sensitive update, check that all votes are cast by a sensitive voting certificate.
-- In both cases, check that all signatures are verifiable, and no superfluous signatures are attached.
+- Check that all signatures are verifiable, and no superfluous signatures are attached.
```
knmeynell: I don't think a comma is needed here.

knmeynell (changes requested):

draft-dekater-scion-pki.md
```suggestion
To fulfill the above requirements, which in fact apply well to inter-domain networking, SCION introduces the concept of **Isolation Domains**. An Isolation Domain (ISD) is a building block to support heterogeneous trust while achieving high availability and scalability in {{I-D.dekater-scion-controlplane}}. It consists of a logical grouping of SCION ASes that share a uniform trust environment (i.e. a common jurisdiction).
```

draft-dekater-scion-pki.md
```suggestion
Authentication in SCION is based on X.509 certificates that bind identifiers to public keys and carry digital signatures that are verified by roots of trust. SCION allows each ISD to define its own set of trust roots, along with the policy governing their use. Such scoping of trust roots within an ISD improves security as compromise of a private key associated with a trust root cannot be used to forge a certificate outside the ISD. An ISD's trust roots and policy are encoded in the TRC, which has a version number, a list of public keys that serves as root of trust for various purposes, and a voting quorum governing the number of signatures required to update TRCs. The TRC serves as a way to bootstrap all authentication within SCION. Additionally, TRC versioning is used to efficiently revoke compromised roots of trust.
```

draft-dekater-scion-pki.md
```suggestion
- **Core ASes**: They are a distinct set of ASes in the SCION Control Plane. For each ISD, the core ASes are listed in the TRC and each core AS has links to the other core ASes (in the same or in different ISDs).
```

draft-dekater-scion-pki.md
```suggestion
- Check that all signatures are verifiable and no superfluous signatures are attached.
```

draft-dekater-scion-pki.md
```suggestion
- **Voting ASes**: They may sign TRC updates. The process of appending a signature to a new TRC is called "casting a vote", and the designated ASes that hold the private keys to sign a TRC update are "voting ASes".
```

draft-dekater-scion-pki.md
```suggestion
This extension MUST be marked as non-critical. Implementations MUST return an error if the extension is not present AND the certificate is not self-signed.
```

draft-dekater-scion-pki.md
```suggestion
This extension MUST be marked as non-critical. Implementations MUST return an error if the extension is not present.
```

draft-dekater-scion-pki.md
```suggestion
This section describes the rules that apply to updating a TRC in regard to the payload information contained in the TRC. Some rules are valid for both update types whilst some only apply to a regular or a sensitive TRC update. Based on the type of update, different sets of voters are needed to create a verifiable TRC update and the corresponding voting (signing) process is also described. Finally, this section describes checks to verify a newly issued TRC.
```

draft-dekater-scion-pki.md
nicorusti: I have to merge this to submit a new draft version and keep the ISE process going. I left a note about this SHOULD in https://github.com/scionassociation/scion-cppki_I-D/issues/132

draft-dekater-scion-pki.md
nicorusti: He commented via email: SGTM

draft-dekater-scion-pki.md
nicorusti: Dominik commented: 
MUST is good

Changing this in #135

PR #135, opened 2026-04-24
Nic review - part 2

### LINKED ISSUE #122: Final review Nico

### LINKED ISSUE #132: Review use of normative language
- [x] Change the two SHOULD to MUST in this line in Table? 
`pathLenConstraint | SHOULD be set to "1", MUST be marked as "critical" | SHOULD be set to "0"`
We use SHOULD as per RFC 5280, but our code is enforcing this (see  Leftover from https://github.com/scionassociation/scion-cppki_I-D/pull/123#discussion_r3088049508)
-  [x] Check that all pending comments in https://github.com/scionassociation/scion-cppki_I-D/pull/123 are addressed
- [ ] Do we need normative language in the appendix (ceremony)? 
Appendixes are often considered informative unless explicitly stated otherwise. If the signing ceremony procedure is a requirement for a valid SCION deployment, having normative language there is fine, but it should be clearly stated at the beginning of the appendix if the entire process is normative.

nicorusti:
Resolves #132 
Resolves #122 
 ...

Diff [with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.txt&url_2=https://scionassociation.github.io/scion-cppki_I-D/nic_review_2/draft-dekater-scion-pki.txt)

### DIFF
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -548,7 +548,7 @@ The attributes of the `keyUsage` extension define possible ways of using the pub
 
 Other attributes are not used.
 
-If a certificate’s public key is used to verify the signature of a control plane payload (`digitalSignature` attribute), it MUST be possible to trace back the private key used to sign the certificate. This is done by referencing the ISD-AS and the subject key identifier (via the `subjectKeyIdentifier` extension). For more information about the `subjectKeyIdentifier` extension (see [](#subject-key-id-ext)).
+If a relying party uses the certificate’s public key to verify the signature of a control plane payload (`digitalSignature` attribute), the relying party MUST be able to trace back the private key used to sign the certificate. This is done by referencing the ISD-AS and the subject key identifier (via the `subjectKeyIdentifier` extension). For more information about the `subjectKeyIdentifier` extension (see [](#subject-key-id-ext)).
 
 When present, this extension SHOULD be marked as critical.
 
@@ -1109,7 +1109,7 @@ The steps REQUIRED to issue a new AS certificate are the following:
 3. The CA uses its CA key and the CSR to create the new AS certificate.
 4. The CA sends the AS certificate back to the AS.
 
-When an AS joins an ISD, the first CSR is sent out of band to one of the CAs as part of the formalities to join the ISD. Subsequent certificate renewals MAY be automated and can leverage the control plane communication infrastructure (see {{I-D.dekater-scion-controlplane}}, section "Renewal of Cryptographic Material").
+When an AS joins an ISD, it sends the first CSR out of band to one of the CAs as part of the formalities to join the ISD. Subsequent certificate renewals MAY be automated and can leverage the control plane communication infrastructure (see {{I-D.dekater-scion-controlplane}}, section "Renewal of Cryptographic Material").
 
 # Deployment Considerations
 
@@ -1386,6 +1386,7 @@ Changes made to drafts since ISE submission. This section is to be removed befor
 ## draft-dekater-scion-pki-13
 {:numbered="false"}
 
+- remove trust Hierarchy subsection and redundant code block
 - Certificate validity recommendations: align to current practice
 - `authorityKeyIdentifier` Extension: clarify support for `authorityCertIssuer` and `authorityCertSerialNumber` attributes
 
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -741,7 +741,7 @@ Note that a trust reset represents a special use case where a new base TRC is cr
 
 ### `votes` {#votes}
 
-The `votes` field contains a sequence of indices referencing the voting certificates in the predecessor TRC. If index i is part of the `votes` field, then the voting certificate at position i in the `certificates` sequence of the predecessor TRC casted a vote on the successor TRC. For more information on the `certificates` sequence, see [](#cert).
+The `votes` field contains a sequence of indices referencing the voting certificates in the predecessor TRC. If index i is part of the `votes` field, then the voting certificate at position i in the `certificates` sequence of the predecessor TRC casted a vote on the successor TRC. The index is 0-based, meaning that 0 represents the first element. For more information on the `certificates` sequence, see [](#cert).
 
 In a base TRC, the `votes` sequence MUST be empty.
 Every entry in the `votes` sequence MUST be unique.
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -485,13 +485,13 @@ In addition to the attributes described in {{RFC5280}} section 4.1.2.4, SCION im
 
 The `id-at-ia` attribute identifies the SCION ISD and AS numbers. Its object identifier is defined in [](#cert-asn1).
 
-The string representation of the `ISD-AS number` attribute MUST follow the formatting defined in {{I-D.dekater-scion-controlplane}}, section "Text Representation" where AS numbers in the lower 32-bit range are represented in decimal notation, and others in hexadecimal notation.
+The `id-at-ia` attribute MUST be included in the `issuer` and `subject` fields of root, issuing CA, and AS certificates. It SHOULD be included in voting certificates.
+In issuing CA certificates, the presence of this attribute ensures that the Control Plane can identify from which AS to retrieve the intermediate CA certificate, thereby avoiding circular dependencies.
 
-Voting AS and issuing CA certificates MUST include the `ISD-AS number` attribute exactly once in the distinguished name of the certificate issuer or owner, specified in the `issuer` or `subject` field respectively. Implementations MUST NOT create nor successfully verify certificates whose `issuer` and `subject` fields do not include the ISD-AS number at all, or include it more than once.
+When present, the `id-at-ia` attribute MUST appear exactly once in a given distinguished name (DN), and implementations MUST reject certificates if the `id-at-ia` appears more than once.
 
-For issuing CA certificates, the inclusion of the `ISD-AS number` ensures the Control Plane knows from which AS to retrieve the certificate, thereby avoiding circular dependencies.
+The string representation of the `id-at-ia` attribute MUST follow the formatting defined in {{I-D.dekater-scion-controlplane}}, section "Text Representation" where AS numbers in the lower 32-bit range are represented in decimal notation, and others in hexadecimal notation.
 
-Voting-only certificates are not required to include the `ISD-AS number` attribute in their distinguished name.
 
 ### `validity`
 
@@ -540,8 +540,6 @@ To ensure deterministic matching, the authorityKeyIdentifier attributes are stri
 
 - `keyIdentifier`: MUST be included.
 - `authorityCertIssuer` & `authorityCertSerialNumber`: MUST NOT be included. Implementations MUST return an error if either is present.
-
-SCION implementations MAY also support the use of the `authorityCertIssuer` and `authorityCertSerialNumber` attributes. However, if these attributes are set and support for them is missing, implementations SHOULD error out.
 
 This extension MUST be marked as non-critical. Implementations MUST return an error if the extension is not present AND the certificate is not self-signed.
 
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-pki.md
nicorusti: > if these attributes are set and support for them is missing, implementations SHOULD error out. 

What does "error out" mean? Reject the certificate? If so, this clashes with the  earlier MAY: an implementation would reject a cert with an attribute that we say is optional.

P.S.: This wording was introduced in a very early [draft version](https://github.com/scionassociation/scion-cppki_I-D/commit/6db6cc71211767996eac23814968bef0d3368d03#diff-46a3cbbb8fca7cc808ffd598a474834aeef4be1c1364b16a354475c27c499dacL603) and it comes straight from [docs](https://docs.scion.org/en/latest/cryptography/certificates.html#authority-key-identifier-extension), I believe we should revise it.

draft-dekater-scion-pki.md
```diff
@@ -534,17 +534,12 @@ The following sections describe the SCION-specifics in regard to these extension
...
 
-The `authorityKeyIdentifier` extension provides three attributes to specify the public key:
-
-- `keyIdentifier`
-- `authorityCertIssuer`
-- `authorityCertSerialNumber`
-
-In SCION, using the `keyIdentifier` attribute is the preferred way to specify the `authorityKeyIdentifier` extension.
+- `keyIdentifier`: MUST be included.
+- `authorityCertIssuer` & `authorityCertSerialNumber`: MUST NOT be included. Implementations MUST return an error if either is present.
```
nicorusti, line 511: As far as I understand, using `authorityCertIssuer` & `authorityCertSerialNumber` is not supported in scionproto. If I'm right, then the old wording was incorrect. I therefore reworded to mention that implementations MUST only use `keyIdentifier` 
Correct @oncilla? 

I realized this section was weird because of this old wording:

> if these attributes are set and support for them is missing, implementations SHOULD error out. 

What does "error out" mean? Reject the certificate? If so, this clashes with the  earlier MAY: an implementation would reject a cert with an attribute that we say is optional.

P.S.: This wording was introduced in a very early [draft version](https://github.com/scionassociation/scion-cppki_I-D/commit/6db6cc71211767996eac23814968bef0d3368d03#diff-46a3cbbb8fca7cc808ffd598a474834aeef4be1c1364b16a354475c27c499dacL603) and it comes straight from [docs](https://docs.scion.org/en/latest/cryptography/certificates.html#authority-key-identifier-extension), this is why I think we have to change wording.

draft-dekater-scion-pki.md
nicorusti: What is a voting-only certificate?  And these two sentences seem conflicting:
> Voting AS and issuing CA certificates MUST include the `ISD-AS number` attribute

> Voting-only certificates are not required to include the `ISD-AS number` attribute 

The first sentence was introduced [here](https://github.com/scionassociation/scion-cppki_I-D/commit/a4fa4628a2b9eadbc0e88053431dc884c0ad86a6), specifically in https://github.com/scionassociation/scion-cppki_I-D/pull/108/changes/b4624299438c05c18c86e5f6748738691acf7e56

nicorusti: This topic was discussed in [Slack](https://scionproto.slack.com/archives/C8ADA9CEP/p1740057439410749)

nicorusti: I reworded this section following Dominik's clarification.

draft-dekater-scion-pki.md
```diff
@@ -485,13 +485,13 @@ In addition to the attributes described in {{RFC5280}} section 4.1.2.4, SCION im
 
 The `id-at-ia` attribute identifies the SCION ISD and AS numbers. Its object identifier is defined in [](#cert-asn1).
 
-The string representation of the `ISD-AS number` attribute MUST follow the formatting defined in {{I-D.dekater-scion-controlplane}}, section "Text Representation" where AS numbers in the lower 32-bit range are represented in decimal notation, and others in hexadecimal notation.
+The `id-at-ia` attribute MUST be included in the `issuer` and `subject` fields of root, issuing CA, and AS certificates. It SHOULD be included in voting certificates.
+In issuing CA certificates, the presence of this attribute ensures that the Control Plane can identify from which AS to retrieve the intermediate CA certificate, thereby avoiding circular dependencies.
```
oncilla, line 458: > In issuing CA certificates, the presence of this attribute ensures that the Control Plane can identify from which AS to retrieve the intermediate CA certificate, thereby avoiding circular dependencies.

That sentence does not make any sense IMO, and should be dropped.

draft-dekater-scion-pki.md
```diff
@@ -368,15 +368,15 @@ The RECOMMENDED maximum validity period of a sensitive voting certificate is 5 y
...
+| CP root certificate    | C<sub>root</sub>  | SK<sub>root</sub>                        | PK<sub>root</sub>                                       | 5 years      |
+| CP issuing CA certificate      | C<sub>CA</sub>    | SK<sub>root</sub>                        | PK<sub>CA</sub>                                 | 15 days (3)  |
 | CP AS certificate      | C<sub>AS</sub>    | SK<sub>CA</sub>                          | PK<sub>AS</sub>                                         | 3 days       |
 {: #table-3 title="Certificates"}
 
 (1) Multiple signatures and certificates of each type MAY be included in a TRC.<br>
 (2) Recommended maximum validity period. Note that initial AS certificates may have a longer validity (e.g. 10-30 days) to allow for enough time for deployment.<br>
-(3) A validity of 11 days with 4 days overlap between two issuing CA certificates is RECOMMENDED to enable the best possible operational procedures when performing a issuing CA certificate rollover.
+(3) A validity of 15 days with 4 days overlap between two issuing CA certificates is RECOMMENDED to enable the best possible operational procedures when performing a issuing CA certificate rollover.
```
oncilla: > A validity of 15 days with 4 days overlap

What we do in practice is to create a new intermediate CA every Monday at the same time, hence the overlap is 8(?) days.

draft-dekater-scion-pki.md
nicorusti: This sentence is meant to explain the reason why you MUST include the ISD-AS attribute sometimes. The reason is to let relying parties discover where to go and fetch the right certificate (therefore avoiding circular dependencies with cryptography). I agree that the old wording is weird. How about something like: 
```suggestion
The presence of this attribute allows relying parties to identify the AS that operates the issuing CA, facilitating the discovery and retrieval of the intermediate CA certificate during chain validation.
```

Let me know what you think @oncilla. I don't think this is absolutely necessary, so if you  are still not convinced, I'll drop the sentence.

draft-dekater-scion-pki.md
nicorusti: Ok changed to 8 days [f52cbdb](https://github.com/scionassociation/scion-cppki_I-D/pull/135/commits/f52cbdb3fff6e46cf86d3363a3b67bb092c04d49)

draft-dekater-scion-pki.md
```diff
@@ -748,7 +741,7 @@ Note that a trust reset represents a special use case where a new base TRC is cr
 
 ### `votes` {#votes}
 
-The `votes` field contains a sequence of indices referencing the voting certificates in the predecessor TRC. If index i is part of the `votes` field, then the voting certificate at position i in the `certificates` sequence of the predecessor TRC casted a vote on the successor TRC. For more information on the `certificates` sequence, see [](#cert).
+The `votes` field contains a sequence of indices referencing the voting certificates in the predecessor TRC. If index i is part of the `votes` field, then the voting certificate at position i in the `certificates` sequence of the predecessor TRC casted a vote on the successor TRC. The index is 0-based, meaning that 0 represents the first element. For more information on the `certificates` sequence, see [](#cert).
```
nicorusti, line 713: I feel this was not clear in the earlier text.  @oncilla Looks good?

nicorusti (commented):
@oncilla could you look at some of my comments, and eventually at the overall diff?

knmeynell (approved):

nicorusti:
@oncilla what do you think about comments in here? 

I have to merge this PR so that others an review, but the points above remain open

draft-dekater-scion-pki.md
oncilla: > As far as I understand, using authorityCertIssuer & authorityCertSerialNumber is not supported in scionproto.

I don't think this is true, no.

While we will never generate the authorityCertIssuer/authorityCertSerialNumber in with the Go code (I cannot attest to the libc implementation, but unlikely they would), the code is still able to deal with certificate chains that have this set without a authorityKeyIdentifier. Essentially, the verification code ignores these paramaters and searches for potential parent certificates:

For reference: https://github.com/golang/go/blob/922abf576dfcd6e836a936b5fcc636e9e572b591/src/crypto/x509/cert_pool.go#L153-L155

So, we can add this additional restriction (which also is pretty in-line with WebPKI), but currently it is not enforced. We would need to tighten down the verification logic.

oncilla:
I cannot comment on this directly: https://github.com/scionassociation/scion-cppki_I-D/pull/135#discussion_r3168700487
So I add the comment here.

> The presence of this attribute allows relying parties to identify the AS that operates the issuing CA, facilitating the discovery and retrieval of the intermediate CA certificate during chain validation.

I think this is technically not correct because the intermediate CA certificate is always  distributed together with the AS certificate in a certificate chain. The only thing we need to match is the root certificate, but that can be done based on the Issuer Subject and AKID. 

It is pretty hard to find a motivation why we have this restriction because it is not technically necessary to have. I think it was necessary before the move to X.509 certificates, and we never reconsidered if this is actually necessary.

What is "necessary" is that the CA is a routable entity in the SCION network to support auto-renewal. It also makes sense to include the ISD-AS of that entity in the subject of the intermediate certificate. However, the whole system would still function afaik if the intermediate CA certificate did not have the ISD-AS number. (excluding our validation logic that enforces it of course). In the control plane, we don't really rely on that information.

For the auto-renewal, we use the ISD-AS of the roots in the TRC to discover potential fallback ISD-ASes when the renewal fails with the primary. 

If not requested, I would just leave the motivation open, and state it as a requirement without it.

draft-dekater-scion-pki.md
oncilla: WFM.

nicorusti:
> If not requested, I would just leave the motivation open, and state it as a requirement without it. 

@oncilla ok, I dropped the sentence in https://github.com/scionassociation/scion-cppki_I-D/pull/147 

Could you also chime in on other comments in this PR?

draft-dekater-scion-pki.md
nicorusti: So you are fine with the current wording, right? 
Screenshot 2026-05-22 at 15 41 25

nicorusti: I see you wrote in #149: 

> As also mentioned elsewhere, currently this is not enforced by our implementation. I am fine however to add this, as we are not setting these fields in our current implementations.

PR #137, opened 2026-04-30
TRC description: mandate normalization, introduce language tag and localizedDescriptions

### LINKED ISSUE #136: Language tags in TRC description (RFC 5198) and UTF-8 normalization
From Alexey Melnikov: 

3.2.10. description

    The description field contains a UTF-8 encoded string that describes
    the ISD. It SHOULD NOT be empty. The description MUST be in English
    and MAY additionally contain information in other languages.

If information in other languages is conveyed, it should be tagged with language tags in some form, see RFC 5646.

Also, if one suggests using UTF-8 for human readable text, one should generally use (reference) something like RFC 5198.

nicorusti:
Resolves #136 

Diff[ with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.txt&url_2=https://scionassociation.github.io/scion-cppki_I-D/description_encoding/draft-dekater-scion-pki.txt)

[Slack discussion](https://github.com/scionassociation/scion-cppki_I-D/pull/137)

TODO: 
- [ ] Ensure latest trc.asn1 is copy pasted to appendix before merging


draft-dekater-scion-pki.md
```diff
@@ -783,8 +785,8 @@ As with core ASes, assigning or revoking authoritative status is performed by ad
 
 ### `description` {#description}
 
-The `description` field contains a UTF-8 encoded string that describes the ISD. It SHOULD NOT be empty.
-The description MUST be in English and MAY additionally contain information in other languages.
+The `description` field contains a UTF-8 encoded string that describes the ISD. The text SHOULD be formatted in accordance with "Net-Unicode" {{RFC5198}} to ensure consistent normalization. It SHOULD NOT be empty.
```
oncilla: We can make it a MUST if you want. All existing TRCs are in that form

```suggestion
The `description` field contains a UTF-8 encoded string that describes the ISD. The text MUST be formatted in accordance with "Net-Unicode" {{RFC5198}} to ensure consistent normalization. It SHOULD NOT be empty.
```

draft-dekater-scion-pki.md
```diff
@@ -783,8 +785,9 @@ As with core ASes, assigning or revoking authoritative status is performed by ad
 
 ### `description` {#description}
 
-The `description` field contains a UTF-8 encoded string that describes the ISD. It SHOULD NOT be empty.
-The description MUST be in English and MAY additionally contain information in other languages.
+The `description` field contains a UTF-8 encoded string that describes the ISD. The text MUST be formatted in accordance with "Net-Unicode" {{RFC5198}} to ensure consistent normalization. It SHOULD NOT be empty.
+The description MUST be in English. If multiple languages are provided, each language section SHOULD be identified by a language tags as defined in {{BCP47}}.
+When using language tags, each tag SHOULD be enclosed in square brackets (e.g., [en], [de-CH]) at the start of its respective section.
```
nicorusti: @oncilla how about this?

draft-dekater-scion-pki.md
```diff
@@ -745,8 +747,9 @@ As with core ASes, assigning or revoking authoritative status is performed by ad
 
 ### `description` {#description}
 
-The `description` field contains a UTF-8 encoded string that describes the ISD. It SHOULD NOT be empty.
-The description MUST be in English and MAY additionally contain information in other languages.
+The `description` field contains a UTF-8 encoded string that describes the ISD. The text MUST be formatted in accordance with "Net-Unicode" {{RFC5198}} to ensure consistent normalization. It SHOULD NOT be empty.
+The description MUST be in English. If multiple languages are provided, each language section SHOULD be identified by a language tags as defined in {{BCP47}}.
```
oncilla: I don't think this can be a MUST for English. We have examples of TRCs that are not in English already.
E.g. ISD70, ISD72, ISD 76. It also feels a bit like an overreach. We want SCION to be adopted across the world. 

There is a logical inconsistency, fwiw. If the description is English, it cannot have multiple languages.

draft-dekater-scion-pki.md
oncilla: I don't think this works well, and feels like an after thought. It also makes all our TRCs go against the SHOULD recommendation.

How about the following:
We add a backward/forward compatible additional field:
`descriptions` (optional, at the end for DER, with explicit tag) that is a list of description and language.
Multi-language TRCs should use the descriptions field. No further restrictions on the description field itself.
(We could even go further and say all TRCs should use the descriptions field rather than description)

I need to double check that this is actually backward and forward compatible.

For inspiration: `LogotypeImageInfo` https://datatracker.ietf.org/doc/html/rfc9399#name-loyalty-logotype

Additionally, we can add a descriptionLanguageTag similar to: `taTitleLangTag` https://datatracker.ietf.org/doc/html/rfc5914#section-2.4

draft-dekater-scion-pki.md
nicorusti: > The description MUST be in English  

This comes from -00 and likely comes from some old docs. I see your point, let's drop it

draft-dekater-scion-pki.md
nicorusti: I updated the text and ASN.1 module to introduce two new optional texts. What do you think?

asn1/trc.asn1
```diff
@@ -1,36 +1,43 @@
...
 TRCValidity ::= SEQUENCE {
     notBefore          GeneralizedTime,
     notAfter           GeneralizedTime
 }
 
+LocalizedText ::= SEQUENCE {
+    languageTag        VisibleString,
+    text               UTF8String (SIZE (0..8192))
+}
+
```
oncilla:
```suggestion
LocalizedText ::= SEQUENCE {
    language        PrintableString,
    content         UTF8String (SIZE (0..8192))
}
```

I think we can tighten down the string type even further.
AFAIU, this PrintableString is the most restrictive option that can still represent the required characters

Also, I think we can remove some redundancy from the text

asn1/trc.asn1
```diff
@@ -1,36 +1,43 @@
...
+    gracePeriod           INTEGER,
+    noTrustReset          BOOLEAN,
+    votes                 SEQUENCE SIZE (0..2047) OF INTEGER (0..4095),
+    votingQuorum          INTEGER (1..2047),
+    coreASes              SEQUENCE OF ASN,
+    authoritativeASes     SEQUENCE OF ASN,
+    description           UTF8String (SIZE (1..8192)) OPTIONAL,
+    certificates          SEQUENCE SIZE (1..4095) OF Certificate,
+    	localizedDescriptions [0] SEQUENCE SIZE (1..MAX) OF LocalizedText OPTIONAL,
+    descriptionLangTag    [1] VisibleString OPTIONAL
```
oncilla:
```suggestion
    descriptionLanguage    [1] PrintableString OPTIONAL
```

oncilla (commented):
Nit: i would change the field names slightly

nicorusti:
Hi @oncilla,  I incorporated your suggestions in the PR.  Good to hear that you like the overall approach, and thanks for posting your comments.. I also sometimes forget to click publish…

> Did you already check that the existing TRCs are compatible with this ASN.1 module?

With the updated ASN.1 module here, I can successfully decode various TRCs in ASN.1 studio: 
Screenshot 2026-05-15 at 15 16 38


In general looks good to me, but I need to double check how existing code will behave if a TRC with these new fields is encountered.
How long does it take? 
I would merge the PR if we don’t expect other changes, but  leave Issue #[136](https://github.com/scionassociation/scion-cppki_I-D/issues/136)  open until you can get this tested. If it helps, I dropped[ a TRC ](https://github.com/scionassociation/scion-cppki_I-D/blob/46e6b9fcc59c42cde4af0b6e08d40d6f44f0d7a0/asn1/trcs/ISD71_trc_4.pem_multilang.der)with the extended description. it would be good to check how this affects signing of the TRC. This way you can take your time. We anyways have at least a few weeks before the draft goes to publication, since Eliot will also do a final review on his side.

If you feel this change needs more community input, I can post it on scionproto, but I’m not sure there are more people understanding PKI as deep as you.

oncilla:
> Hi @oncilla, I incorporated your suggestions in the PR. Good to hear that you like the overall approach, and thanks for posting your comments.. I also sometimes forget to click publish…
> 
> > Did you already check that the existing TRCs are compatible with this ASN.1 module?
> 
> With the updated ASN.1 module here, I can successfully decode various TRCs in ASN.1 studio: Screenshot 2026-05-15 at 15 16 38

Cool. Thanks for checking!

> In general looks good to me, but I need to double check how existing code will behave if a TRC with these new fields is encountered. How long does it take? I would merge the PR if we don’t expect other changes, but leave Issue #[136](https://github.com/scionassociation/scion-cppki_I-D/issues/136) open until you can get this tested. If it helps, I dropped[ a TRC ](https://github.com/scionassociation/scion-cppki_I-D/blob/46e6b9fcc59c42cde4af0b6e08d40d6f44f0d7a0/asn1/trcs/ISD71_trc_4.pem_multilang.der)with the extended description. it would be good to check how this affects signing of the TRC. This way you can take your time. We anyways have at least a few weeks before the draft goes to publication, since Eliot will also do a final review on his side.

I created a draft PR with what I did to check: https://github.com/scionproto/scion/pull/4927
As far as I can tell, the change is forward and backward compatible. So all good from my side!

> If you feel this change needs more community input, I can post it on scionproto, but I’m not sure there are more people understanding PKI as deep as you.

It's probably good form to announce this in the community. I don't think there is going to be much input though.

asn1/trc.asn1
```diff
@@ -1,36 +1,42 @@
...
         };
 
 TRCValidity ::= SEQUENCE {
     notBefore          GeneralizedTime,
     notAfter           GeneralizedTime
 }
 
+LocalizedText ::= SEQUENCE {
+    language        PrintableString,
+    content         UTF8String (SIZE (0..8192))
```
tzaeschke:
```suggestion
    content         UTF8String (SIZE (1..8192))
```

asn1/trc.asn1
```diff
@@ -1,36 +1,42 @@
...
+    validity              TRCValidity,
+    gracePeriod           INTEGER,
+    noTrustReset          BOOLEAN,
+    votes                 SEQUENCE SIZE (0..2047) OF INTEGER (0..4095),
+    votingQuorum          INTEGER (1..2047),
+    coreASes              SEQUENCE OF ASN,
+    authoritativeASes     SEQUENCE OF ASN,
+    description           UTF8String (SIZE (1..8192)) OPTIONAL,
+    certificates          SEQUENCE SIZE (1..4095) OF Certificate,
+    	localizedDescriptions [0] SEQUENCE SIZE (1..MAX) OF LocalizedText OPTIONAL,
```
tzaeschke: Instead of MAX, I suggest using a very low number here, e.g. `5`.

I also wonder, do we really need 8KB?
How about limiting it to 256 bytes?
Same for the localizations. Also, localizations don't need to have the same length as the main description...?

The problem is really that this can be exploited to create really large TRC that could even be used as an attack (as discussed in the meeting yesterday).

draft-dekater-scion-pki.md
```diff
@@ -745,9 +747,10 @@ As with core ASes, assigning or revoking authoritative status is performed by ad
 
 ### `description` {#description}
 
-The `description` field contains a UTF-8 encoded string that describes the ISD. It SHOULD NOT be empty.
-The description MUST be in English and MAY additionally contain information in other languages.
+The `description` field contains a UTF-8 encoded string that describes the ISD. The text MUST be formatted in accordance with "Net-Unicode" {{RFC5198}} to ensure consistent normalization. If present, it MUST NOT be empty.
```
tzaeschke: This is already implied by the ASN.1 description, which specifies a minimum size of 1. Do we need an additional MUST here?
Id suggest removing "If present, it MUST NOT be empty."

draft-dekater-scion-pki.md
```diff
@@ -745,9 +747,10 @@ As with core ASes, assigning or revoking authoritative status is performed by ad
...
+When this field contains languages other than English, the corresponding language tag SHOULD explicitly identified in the `descriptionLanguage` field (see ()[#langtag]).
```
```suggestion
When this field contains languages other than English, the corresponding language SHOULD be identified explicitly in the `descriptionLanguage` field (see ()[#langtag]).
```

asn1/trc.asn1
tzaeschke:
```suggestion
    content         UTF8String (SIZE (1..256))
```
See discussion below on `description` field: We should probably avoid bloated TRC.

asn1/trc.asn1
tzaeschke: The size concern also goes for other fields, do we really need 2000 votes or 4000 certificates?

draft-dekater-scion-pki.md
```diff
@@ -763,6 +766,18 @@ A certificate that is no control plane root or voting certificate MUST NOT be in
 
 A certificate's type (voting or root) is specified in the `extKeyUsage` extension of the certificate, by means of the SCION-specific attributes `id-kp-regular`, `id-kp-sensitive`, and `id-kp-root`, respectively. For more information, see [](#ext-key-usage-ext).
 
+### `localizedDescriptions` {#description-multilang}
+
+The `localizedDescriptions` field provides an optional mechanism for including multilingual descriptions.
+It consists of a sequence of `LocalizedText` structures, each containing:
+
+- `languageTag`: specifies the description's language. It MUST use language tags according to {{BCP47}}.
+- `text`: contains the localized description. It MUST be formatted in accordance with "Net-Unicode" {{RFC5198}}.
```
```suggestion
- `language`: specifies the description's language. It MUST use language tags according to {{BCP47}}.
- `content`: contains the localized description. It MUST be formatted in accordance with "Net-Unicode" {{RFC5198}}.
```

draft-dekater-scion-pki.md
```diff
@@ -745,9 +747,10 @@ As with core ASes, assigning or revoking authoritative status is performed by ad
 
 ### `description` {#description}
 
-The `description` field contains a UTF-8 encoded string that describes the ISD. It SHOULD NOT be empty.
-The description MUST be in English and MAY additionally contain information in other languages.
+The `description` field contains a UTF-8 encoded string that describes the ISD. The text MUST be formatted in accordance with "Net-Unicode" {{RFC5198}} to ensure consistent normalization. If present, it MUST NOT be empty.
+When this field contains languages other than English, the corresponding language tag SHOULD explicitly identified in the `descriptionLanguage` field (see ()[#langtag]).
 
+Multi-language TRCs SHOULD use the `localizedDescriptions` field instead. Either the `description` or the `localizedDescriptions`field MUST be present.
```
```suggestion
Multi-language TRCs SHOULD use the `localizedDescriptions` field instead of the `description` field. Either the `description` or the `localizedDescriptions`field MUST be present.
```

It is a bit unclear what "instead" refers to. Does it mean that one should use ether "description" or "localizedDescriptions", but not both?

draft-dekater-scion-pki.md
```diff
@@ -1169,18 +1184,24 @@ TRCValidity ::= SEQUENCE {
     notAfter           GeneralizedTime
 }
 
+LocalizedText ::= SEQUENCE {
+    language        PrintableString,
+    content         UTF8String (SIZE (0..8192))
```
tzaeschke: See discussion on trc.asn1 above.

draft-dekater-scion-pki.md
```diff
@@ -1169,18 +1184,24 @@ TRCValidity ::= SEQUENCE {
...
+    validity              TRCValidity,
+    gracePeriod           INTEGER,
+    noTrustReset          BOOLEAN,
+    votes                 SEQUENCE SIZE (0..2047) OF INTEGER (0..4095),
+    votingQuorum          INTEGER (1..2047),
+    coreASes              SEQUENCE OF ASN,
+    authoritativeASes     SEQUENCE OF ASN,
+    description           UTF8String (SIZE (1..8192)) OPTIONAL,
+    certificates          SEQUENCE SIZE (1..4095) OF Certificate,
+    localizedDescriptions [0] SEQUENCE SIZE (1..MAX) OF LocalizedText OPTIONAL,
```
tzaeschke, line 1199: See discussion on trc.asn1 above.

asn1/trc.asn1
tzaeschke:
```suggestion
   localizedDescriptions [0] SEQUENCE SIZE (1..MAX) OF LocalizedText OPTIONAL,
```

draft-dekater-scion-pki.md
nicorusti: A field containing just a whitespace would be parsed and OK according to the ASN.1 module I don't think that is the point, so I recommend leave this wording where it is.

draft-dekater-scion-pki.md
nicorusti: Le't work on the .asn1 file firt, then I copy paste it into the draft at the end

draft-dekater-scion-pki.md
nicorusti: Or at least leaving a SHOULD: 
```suggestion
The `description` field contains a UTF-8 encoded string that describes the ISD. The text MUST be formatted in accordance with "Net-Unicode" {{RFC5198}} to ensure consistent normalization. If present, it SHOULD NOT be empty.
```

asn1/trc.asn1
nicorusti: Is TRC size a real threat vector? If yes, is it prevented with hard limits in the spec?

I mean, if you see a 1 GB certificate/TRC, you maybe don't want to open it anyways and I think implementations may have some limits internally. Is it the case?

asn1/trc.asn1
nicorusti: The current SSHN TRC does not fit (even tough it uses multiple languages without tags). 
https://github.com/scionassociation/scion-cppki_I-D/pull/127#discussion_r3065519614

nicorusti: With @oncilla we discussed [here](https://github.com/scionassociation/scion-cppki_I-D/pull/127#discussion_r3071876416) that 4096 or 8192 is an acceptable size.. I would not make it smaller than 4096 (or bare minimum 1024)

nicorusti: We have to review many field limits, I suggest that we merge this one and then just review field sizes in https://github.com/scionassociation/scion-cppki_I-D/pull/153

nicorusti:
> It's probably good form to announce this in the community. I don't think there is going to be much input though. 

FOr the record, this was discussed at TC #10 on 21.05.2026 with no objections. It was also announced on [Slack](https://scionproto.slack.com/archives/C8ADA9CEP/p1779297622605409) and it will be brought to the OS contributor call next week.

asn1/trc.asn1
nicorusti: We have to review many field limits, I suggest that we merge this PR, which is about language tags, and then review field sizes in https://github.com/scionassociation/scion-cppki_I-D/pull/153

draft-dekater-scion-pki.md
tzaeschke: > A field containing just a whitespace would be parsed and OK according to the ASN.1 module I don't think that is the point, so I recommend leave this wording where it is.

I think what you really want to specify here is that the content should be **sensible**, but I am not sure that can specified with normative language.

It is also debatable whether a string with a single character is empty, regardless of whether this is a whitespace or not. If you want to prevent a whitespace, a more precise norm would be: `If present, it MUST NOT contain only whitespaces.` ...?

That would still allow descriptions like `123456`, so I think this cannot be resolved with normative language and I suggest to remove it.

tzaeschke (approved):
I approve, but I think we need to discuss the "NOT empty" further, I think it doesn't express what it is supposed to express (see my comment below)....

draft-dekater-scion-pki.md
nicorusti: I would still encourage deployers to put something in that field.. I don't think we can be the protocol police and avoid that people put 12345 in it, but I think we should at least give some human-language guidance, and I see no harm in having a SHOULD not be empty. It becomes a lot clearer than just reading the 1 in the ASN.1 module

tzaeschke: Okay :+1:

tzaeschke (approved):

nicorusti:
For the record, enclosed the html diff corresponding to this PR
[Diff_ draft-dekater-scion-pki.txt - draft-dekater-scion-pki.txt.html](https://github.com/user-attachments/files/28155534/Diff_.draft-dekater-scion-pki.txt.-.draft-dekater-scion-pki.txt.html)

PR #140, opened 2026-05-01
remove redundant trust as a function section

### LINKED ISSUE #124: 1.6. Trust as a Function --> move close to TRC ceremony?

nicorusti:
Resolves #124

### DIFF
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -189,6 +189,7 @@ The need for trust agility also means that SCION does not by design provide IP p
 
 The Control Plane PKI is organized at an ISD level whereby each ISD can independently specify its own Trust Root Configuration (TRC) and build its own verification chain. Each TRC consists of a collection of signed root certificates which are used to sign issuing CA certificates, which are in turn used to sign AS certificates. The TRC also includes ISD policies that specify, for example, the TRC's usage, validity, and future updates. The so-called **base TRC** constitutes the ISD's trust anchor which is signed during a signing ceremony by the voting ASes and then distributed throughout the ISD.
 
+While it is not necessary that all the ASes of the ISD trust each other, all ASes MUST trust the ISD's core ASes, authoritative ASes, voting ASes, as well as its CA(s).
 
 ### Updates and Trust Resets {#trust-reset}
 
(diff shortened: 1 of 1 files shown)


draft-dekater-scion-pki.md
```diff
@@ -252,7 +253,7 @@ The base TRC constitutes the root of trust within an ISD. {{figure-1}} provides
 ~~~~
 {: #figure-1 title="Chain of trust within an ISD"}
 
-All certificates used in the Control Plane PKI are in X.509 v3 format {{RFC5280}} and additionally the TRC contains self-signed certificates instead of plain public keys. Self-signed certificates have the following advantages over plain public keys: (1) They make the binding between name and public key explicit; and (2) the binding is signed to prove possession of the corresponding private key. The public keys of Voting AS certificates must therefore be explicitly verified during the Signing Ceremony ([](#initial-ceremony)).
+All certificates used in the Control Plane PKI are in X.509 v3 format {{RFC5280}} and additionally the TRC contains self-signed certificates instead of plain public keys. Self-signed certificates have the following advantages over plain public keys: (1) They make the binding between name and public key explicit; and (2) the binding is signed to prove possession of the corresponding private key. The public keys of Voting AS certificates must therefore be explicitly verified during the Signing Ceremony ([](#initial-ceremony)), that is used to bootstrap trust for the initial TRC.
```
knmeynell: I don't think a comma is needed here. 

The public keys of Voting AS certificates must therefore be explicitly verified during the Signing Ceremony ([](#initial-ceremony)) that is used to bootstrap trust for the initial TRC.

knmeynell (changes requested):

draft-dekater-scion-pki.md
```suggestion
All certificates used in the Control Plane PKI are in X.509 v3 format {{RFC5280}} and additionally the TRC contains self-signed certificates instead of plain public keys. Self-signed certificates have the following advantages over plain public keys: (1) They make the binding between name and public key explicit; and (2) the binding is signed to prove possession of the corresponding private key. The public keys of Voting AS certificates must therefore be explicitly verified during the Signing Ceremony ([](#initial-ceremony)) that is used to bootstrap trust for the initial TRC.
```

nicorusti:
@knmeynell feel free to merge of text is ok

Issue #39, opened 2024-08-07
Clarify relationship to RPKI


nicorusti:
Add a couple of sentences to clarify relationship to RPKI

>Date: 	Tue, 6 Aug 2024 23:44:57 +0000
From: 	Michael McBride [<michael.mcbride@futurewei.com>](mailto:michael.mcbride@futurewei.com)
To: 	[rfc-ise@rfc-editor.org](mailto:rfc-ise@rfc-editor.org) [<rfc-ise@rfc-editor.org>](mailto:rfc-ise@rfc-editor.org)
CC: 	'Mike McBride' [<mmcbride7@gmail.com>](mailto:mmcbride7@gmail.com)F. How else can the document be improved?
 
> I’m a believer. It’s a good set of solutions. Using a PKI is a good solution. As is the current RPKI and ROA’s. It might also be helpful, in the scion control plan pki document, to at least acknowledge RPKI. RPKI is widely used now and it could be helpful to understand how scions use of PKI would be similar and different. In the long run it would give operators a more comfortable understanding since most now use RPKI.  
We could add a couple of sentences about this. We have some text in the now expired [draft-rustignoli-panrg-scion-components](https://www.ietf.org/archive/id/draft-rustignoli-panrg-scion-components-03.t): 

We have some text in the now expired [draft-rustignoli-panrg-scion-components](https://www.ietf.org/archive/id/draft-rustignoli-panrg-scion-components-03.t): 
```
   The above-mentioned decoupling also implies that SCION does not
   provide, by design, IP prefix origin validation, which is currently
   provided by RPKI [RFC8210].  As prefix origin validation is outside
   of SCION's scope, IP-to-SCION's coexistence mechanisms (SIAM, SBAS)
   later discussed in Section 3.1 build on top of RPKI for IP origin
   attestation.
   ... 
   These properties would be lost if SCION were to rely on an existing
   PKI (i.e., the web PKI, the RPKI, ...).  For example, if SCION were
   to use the RPKI instead of the CP-PKI, its control plane would lack
   the trust model required to support Isolation Domains.  This is
   because RPKI's trust model follows the same structure as the IP
   allocation hierarchy, where the five RIRs represent the trust roots.
```

knmeynell:
Added reference to why RPKI isn't directly supported by SCION. Did not mention SIAM or SBAS as this appears to be experimental and would need to be further explained somewhere.

PR #146, opened 2026-05-13
Review Chapter 1 - Till and Dominik

### LINKED ISSUE #142: Review Til Chpt 1
Disclaimer: I have little understanding of PKI, so apologies for asking questions that may be too obvious...

### 1.1 Terminology
- General: The ordering seems a bit random, maybe explain it?
- Add references to chapters, e.g. for Trust Reset?

- Authorative AS: the latest TRCs -> "TRC"?  More than one latest TRC?

- Voting quorum: Remove explanation about malicious update. This is a terminology section only.

### 1.3 Trust Model

- This section appears to mix bold font vs italic font for emphasis
- I am not sure about this section, it sounds more like advertisement and bashing of monopoly/oligopoly, maybe tone it down a bit?
 - Remove 2nd and 3rd paragraph
  - Change 4th paragraph to `The SCION trust architecture provides the following properties:`
  - Remove paragraph "Ideally ..."
 - Change sentence "To fulfill..." to `To fulfill the these requirements, SCION introduces the concept of Isolation Domains.`  

- Special mention of the 3rd paragraph (on oligopoly): This sounds like a catchy but over-the-top sales pitch to me, it makes it sound like oligarchy is really worse than monopoly. I think it is catchy at the cost of correctness because it is mixing up two severities/definitions of "failure" here:
  - monopoly single p-o-f: In case of failure, everything is bust
  - oligopoly p-o-f: In case of failure, one can use another root -> some inconvenience

- I am also not entirely sure about these questions:
  - If my parent AS is fraudulent, I have now way of ever finding out. So this is still single point of failure because it can pretend arbitrary ISD topologies + certificates, can't it?
  - If there is only one voting ASes -> single point of failure, right?
  

- If we discuss monopoly and oligopoly at all, we could add following differences:
  - An AS can participate in multiple ISDs. So from an AS perspective, SCION is similar to an oligopoly setup, however, individual TRCs are usually better protected than with today's monopoly-CAs.
  - The actual advantage of SCION is: 
     - In case of failure, the blast radius is a single TRC/ISD
     - Every AS can choose their TRC to be someone they "actually" trust, rather than some entity in a different country/jurisdiction.

- 2nd to last paragraph starting "The TRC also provide trust agility..":
  - Change 
    `This implies that users are free to choose an ISD they believe maintains a uncompromised set of trust roots. ISD members can also decide whether to trust other ISDs and thus transparently define trust relationships between parts of the network.`
    to
    `This implies that ASes are free to choose one or more ISDs they believe maintain a uncompromised set of trust roots. ASes can thus transparently define trust relationships between parts of the network.` 
  - Remove last sentence with comparison to other PKIs, this sounds like unbacked advertisement. 

### 1.4 [Trust Relations within an Isolation Domain](https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.html#name-trust-relations-within-an-i)

- 2nd paragraph, "MUST": Should this be here? This MUST is not really actionable, it is implicit.... replace "MUST" with "implicitly".

#### 1.4.1: 
- Maybe remove this paragraph? It doesn't seem to add anything here. If we keep it, we should explain _why_ there are different types of TRC updates. (I have read Chpt 1 and 2 so far and to me it is not clear yet why there are two types)
- Probably: "If the _base_ TRC has been compromised ..."?
- Explain: why no signing ceremony in case of trust reset? See 1.1.

#### 1.4.2:
- Contradiction: "does not support certificate revocation" but 1.3 states "TRC versioning is used to efficiently revoke compromised roots of trust"...?
- "Instead it relies on the two mechanisms described above and on short-lived certificates."  Which two mechanisms? Unclear what is meant here, are "regular" and "sensitive" mechanisms?

- Now this section describes reasons why the "two mechanisms" are better. Maybe move these reasons to where the two mechanisms are described?
- Bullet 1) This is unclear to me: Are "short-lived" and "revocation lists" are part of the SCION system or a description of the disadvantages of other systems? Obviously SCION uses "short-lived" but it also says that it does not support explicit revocation (which probably refers to revocation lists)...?
- What are CRL/OCSP check? Reference?

### 1.5 [1.5. ](https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.html#section-1.5)[Overview of Certificates, Keys, and Roles](https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.html#name-overview-of-certificates-ke)
- Figure 1: What is the name of the TRC number, e.g. TRC 1, TRC 2, ...? I think it is the "ID"? Maybe clarify to TRC (ID=2) , ...
- Voting ASes: How are they notified of a new TRC to sign? 
- Authorative ASes: Do they get contacted by voting ASes when a TRC update is ready to be distributed?

knmeynell:
Resolves #142 
Contributes to #149

- Removed the paragraphs on monopoly and oligopoly models. I agree it doesn't add anything.
- The mix of bold and italic was correct in the context it's used: bold refers to defined things, whilst italic refers to concepts.
- Removed the reference to RPKI: agree it doesn't add anything although it may have been added FYI for reviewers' benefit.

[Diff with main ](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.txt&url_2=https://scionassociation.github.io/scion-cppki_I-D/Til-feedback/draft-dekater-scion-pki.txt)


draft-dekater-scion-pki.md
```diff
@@ -212,13 +206,13 @@ The base TRC constitutes the root of trust within an ISD. {{figure-1}} provides
 ~~~aasvg
 
                +----------------------------------------+
-               |                 TRC 2                  |
+               |               TRC (iD=2)               |
```
oncilla: shouldn't this be serialNumber, rather than iD?

draft-dekater-scion-pki.md
```diff
@@ -350,7 +344,7 @@ The RECOMMENDED maximum validity period of a sensitive voting certificate is 5 y
 
 ~~~aasvg
 +----------------------------------------------+
-|                    TRC 1                     |
+|                  TRC (id=1)                  |
```
oncilla: ditto (but it is also redundant because initial implies serialNumber == 1)

oncilla (changes requested):

draft-dekater-scion-pki.md
```diff
@@ -161,48 +163,40 @@ SCION relies on three main components:
...
 
-The monopoly model suffers from two main drawbacks: First, all parties must agree on a single root of trust. Secondly, the single root of trust represents a single point of failure, the misuse of which enables the forging of certificates. Its revocation can also result in a kill switch for all the entities it certifies.
-
-The oligopoly model relies on several roots of trust, all equally and completely trusted. However, this is not automatically better: whereas the monopoly model has a single point of failure, the oligopoly model has the drawback of exposing more than one point of failure.
-
-Thus, there is a need for a trust architecture that supports meaningful trust roots in a global environment with inherently distrustful parties. This new trust architecture should provide the following properties:
+The SCION trust architecture allows parties that mutually trust each other to form their own trust domain, and to freely decide whether to trust other trust domains. It therefore provides the following properties
```
nicorusti: I think this text is important to distinguish the PKI's trust model from other PKIs. We can reword it, but I would not delete it, otherwise we are missing some points that motivate SCION having its own PKI.

nicorusti: How about: 

> The monopoly model requires global consensus on a single root of trust, which introduces a single point of failure. The oligopoly model, conversely, utilizes multiple roots of trust that are typically afforded equal authority, introducing multiple potential points of vulnerability, as the compromise of any individual root can undermine the security of the broader ecosystem.

draft-dekater-scion-pki.md
```diff
@@ -161,48 +163,40 @@ SCION relies on three main components:
...
 
-The TRC also provides *trust agility* - enabling users to select the trust roots used to initiate certificate validation. This implies that users are free to choose an ISD they believe maintains an uncompromised set of trust roots. ISD members can also decide whether to trust other ISDs and thus transparently define trust relationships between parts of the network. The SCION trust model therefore, differs from the one provided by other PKI architectures.
-
-The need for trust agility also means that SCION does not by design provide IP prefix origin validation as provided by RPKI {{RFC8210}}. RPKI's trust model is currently reliant on the trust roots provided by the five Regional Internet Registries, and therefore outside of the governance of an ISD.
```
nicorusti, line 187: Why remove this? It is a recurring question and it was also requested by a reviewer
https://github.com/scionassociation/scion-cppki_I-D/issues/39

draft-dekater-scion-pki.md
```diff
@@ -161,48 +163,40 @@ SCION relies on three main components:
...
 
-The TRC also provides *trust agility* - enabling users to select the trust roots used to initiate certificate validation. This implies that users are free to choose an ISD they believe maintains an uncompromised set of trust roots. ISD members can also decide whether to trust other ISDs and thus transparently define trust relationships between parts of the network. The SCION trust model therefore, differs from the one provided by other PKI architectures.
```
nicorusti, line 185: > The TRC also provides *trust agility* - enabling users to select the trust roots used to initiate certificate validation. This implies that ASes are free to choose one or more ISDs they believe maintain an uncompromised set of trust roots. ASes can thus transparently define trust relationships between parts of the network, which differs from trust models provided by other PKI architectures.

This new wording does not work. The old text was clearly talking about users selecting trust roots for a reason. It was conveying that users (or better - endpoints) send traffic through ASes that are only in trusted ISDs (depending on which TRCs are provisioned on the endpoint). I would stick to the old wording here.

> ASes are free to choose one or more ISDs 

Also, in the spec, there is no such thing. An AS in two ISD is two different ISD-ASes tuples, two different CP instances, etc..

draft-dekater-scion-pki.md
```diff
@@ -161,48 +163,40 @@ SCION relies on three main components:
...
 
-While it is not necessary that all the ASes of the ISD trust each other, all ASes MUST trust the ISD's Core ASes, Authoritative ASes, Voting ASes, as well as its CA(s).
+While it is not necessary that all the ASes of the ISD trust each other, all ASes implicitly trust the ISD's Core ASes, Authoritative ASes, Voting ASes, as well as its CA(s).
 
 ### Updates and Trust Resets {#trust-reset}
 
 There are two types of TRC updates: regular and sensitive (see [](#update)). A **regular TRC update** is a periodic re-issuance of the TRC where the entities and policies listed in the TRC remain unchanged, whereas a **sensitive TRC update** is an update that modifies critical aspects of the TRC, such as the set of Core ASes. In both cases the base TRC remains unchanged.
```
nicorusti: Till suggested that this text is redundant. How about shortening it:
```suggestion
There are two types of TRC updates: regular and sensitive. The update type depends on which fields are changed (see [](#update)). In both cases the base TRC remains unchanged.
```

draft-dekater-scion-pki.md
```diff
@@ -161,48 +163,40 @@ SCION relies on three main components:
...
 
-If the ISD's TRC has been compromised, it is necessary for an ISD to re-establish the trust root. This is possible with a process called **trust reset** (if permitted by the ISD's trust policy). In this case, a new base TRC is created.
+If the ISD's base TRC has been compromised, it is necessary for an ISD to re-establish the trust root. This is possible with a process called **trust reset** (if permitted by the ISD's trust policy) when a new base TRC is created.
```
nicorusti: No, a trust reset is necessary if any TRC is compromised, not just a base TRC
```suggestion
If the ISD's TRC has been compromised, it is necessary for an ISD to re-establish the trust root. This is possible with a process called **trust reset** (if permitted by the ISD's trust policy) when a new base TRC is created.
```

draft-dekater-scion-pki.md
```diff
@@ -161,48 +163,40 @@ SCION relies on three main components:
...
 
 
 ### Substitutes to Certificate Revocation {#substitutes-to-revocation}
 
-The Control Plane PKI does not explicitly support certificate revocation. Instead it relies on the two mechanisms described above and on short-lived certificates. This approach constitutes an attractive alternative to a revocation system for the following reasons:
+The Control Plane PKI does not explicitly support certificate revocation. Instead it relies on the regular and sensitive TRC update mechanisms on short-lived certificates. This approach constitutes an alternative to a revocation system for the following reasons:
```
nicorusti: This text does not make sense, how about

```suggestion
The Control Plane PKI does not explicitly support certificate revocation. Instead it relies on the TRC update mechanism and on short-lived certificates. This approach constitutes an alternative to a revocation system for the following reasons:
```

draft-dekater-scion-pki.md
```diff
@@ -161,48 +163,40 @@ SCION relies on three main components:
...
 
 Authentication in SCION is based on X.509 certificates that bind identifiers to public keys and carry digital signatures that are verified by roots of trust. SCION allows each ISD to define its own set of trust roots, along with the policy governing their use. Such scoping of trust roots within an ISD improves security as compromise of a private key associated with a trust root cannot be used to forge a certificate outside the ISD. An ISD's trust roots and policy are encoded in the TRC, which has a version number, a list of public keys that serves as root of trust for various purposes, and a voting quorum governing the number of signatures required to update TRCs. The TRC serves as a way to bootstrap all authentication within SCION. Additionally, TRC versioning is used to efficiently revoke compromised roots of trust.
```
nicorusti: To address till's point: 
> Contradiction: "does not support certificate revocation" but 1.3 states "TRC versioning is used to efficiently revoke compromised roots of trust"...?


```suggestion
Authentication in SCION is based on X.509 certificates that bind identifiers to public keys and carry digital signatures that are verified by roots of trust. SCION allows each ISD to define its own set of trust roots, along with the policy governing their use. Such scoping of trust roots within an ISD improves security as compromise of a private key associated with a trust root cannot be used to forge a certificate outside the ISD. An ISD's trust roots and policy are encoded in the TRC, which has a version number, a list of public keys that serves as root of trust for various purposes, and a voting quorum governing the number of signatures required to update TRCs. The TRC serves as a way to bootstrap all authentication within SCION. Additionally, TRC versioning is used as an alternative to revocation in case of compromised roots of trust.
```

draft-dekater-scion-pki.md
```diff
@@ -161,48 +163,40 @@ SCION relies on three main components:
...
 
 
 ## Trust Relations within an Isolation Domain {#trust-relations}
 
 The Control Plane PKI is organized at an ISD level whereby each ISD can independently specify its own Trust Root Configuration (TRC) and build its own verification chain. Each TRC consists of a collection of signed root certificates which are used to sign issuing CA certificates, which are in turn used to sign AS certificates. The TRC also includes ISD policies that specify, for example, the TRC's usage, validity, and future updates. The so-called **base TRC** constitutes the ISD's trust anchor which is signed during a Signing Ceremony by the Voting ASes and then distributed throughout the ISD.
 
-While it is not necessary that all the ASes of the ISD trust each other, all ASes MUST trust the ISD's Core ASes, Authoritative ASes, Voting ASes, as well as its CA(s).
+While it is not necessary that all the ASes of the ISD trust each other, all ASes implicitly trust the ISD's Core ASes, Authoritative ASes, Voting ASes, as well as its CA(s).
```
nicorusti: @oncilla commented in #149: 

> I wonder if we need to distinguish the level of trust that we need to have in these entities. For the CP-PKI, we do not really need to trust Core or Autoritative ASes. CAs and Voters need to be trusted though.

So how about: 

 
```suggestion
While it is not necessary that all the ASes of the ISD trust each other, within the CP-PKI all ASes implicitly trust the ISD's Voting ASes, as well as its CA(s).
```

nicorusti (changes requested):
@tzaeschke we tried to address your review, but I have some comments on various points. Could you have a look?

draft-dekater-scion-pki.md
nicorusti: Yes 
```suggestion
               |         TRC (serialNumber =2)          |
```

draft-dekater-scion-pki.md
```suggestion
The SCION trust architecture allows parties that mutually trust each other to form their own trust domain, and to freely decide which other trust domains should be trusted. It therefore provides the following properties:
```
I like this test (maybe with some changes if you like them). I don't think there is a need to define what we mean my monopoly/oligopoly if we never use these terms again. Using them just once sounds like we are trying to devalue other approaches, simply by giving them names with a a negative connotation.

draft-dekater-scion-pki.md
tzaeschke: I agree that something got lost. 

> > ASes are free to choose one or more ISDs

> Also, in the spec, there is no such thing. An AS in two ISD is two different ISD-ASes tuples, two different CP instances, etc..

I'm not sure how that contradicts the original statement. I'd say `ASes are free to choose one or more ISDs, subject to available connectivity.`. That they need multiple CS seems more like a detail to me, it's not explicitly mentioned in the CP/DP, but it is also not forbidden. 
However, this is the PKI draft, and on the PKI level, joining multiple ASes is not a problem, or maybe I am mistaken?

Obviously I'm aiming to have this compatible with Private ISDs... :)

draft-dekater-scion-pki.md
```diff
@@ -126,31 +127,32 @@ SCION relies on three main components:
...
+**Control Plane PKI (CP-PKI)**: It is the Public Key Infrastructure upon which SCION's Control Plane relies for the authentication of messages. It is a set of policies, roles, and procedures that are used to manage Trust Root Configurations (TRCs) and certificates.
 
-**Trust Root Configuration (TRC)**: A Trust Root Configuration or TRC is a signed collection of certificates pertaining to an Isolation Domain (ISD). TRCs also contain ISD-specific policies.
+**Core AS**: Each Isolation Domain (ISD) is administered by a set of distinguished SCION Autonomous Systems (ASes) called Core ASes, which are responsible for initiating the path discovery and path construction process (called "beaconing" in SCION). Each ISD has at least one Core AS.
 
-**Authoritative AS**: Authoritative ASes are those ASes in an ISD that always have the latest TRCs of the ISD. As a consequence, Authoritative ASes also start the announcement of a TRC update.
+**Grace Period**: The grace period is an interval during which the previous version of a Trust Root Configuration (TRC) is still considered active after a new version has been published.
```
tzaeschke: I'm now unsure the Grace Period really needs to be here, this seems like oddly specific terminology. Maybe just remove it?

draft-dekater-scion-pki.md
```diff
@@ -126,31 +127,32 @@ SCION relies on three main components:
...
+**Isolation Domain (ISD)**: SCION ASes are organized into logical groups called Isolation Domains or ISDs. Each ISD consists of ASes that span an area with a uniform trust environment (e.g. a common jurisdiction).
 
 **TRC Signing Ceremony**: The ceremony during which the very first base TRC of an ISD, called the initial TRC, is signed. The initial TRC is a special case of the base TRC where the number of the ISD is assigned.
 
 **TRC Update**: A *regular* TRC update is a periodic re-issuance of the TRC where the entities and policies listed in the TRC remain unchanged. A *sensitive* TRC update is an update that modifies critical aspects of the TRC, such as the set of Core ASes. In both cases, the base TRC remains unchanged.
 
-**Voting ASes**: Those ASes within an ISD that may sign TRC updates. The process of appending a signature to a new TRC is called "casting a vote".
+**Trust Reset**: A trust reset is the action of announcing a new base TRC for an existing ISD, to mitigate a compromised TRC.
```
```suggestion
**Trust Reset**: A trust reset is the action of creating and announcing a new base TRC for an existing ISD, to mitigate a compromised TRC.
```

draft-dekater-scion-pki.md
```diff
@@ -161,48 +163,40 @@ SCION relies on three main components:
...
+An ISD is governed by one or multiple ASes, known as the **Voting ASes**. Furthermore, each ISD has a set of ASes that form the ISD core, known as the **Core ASes**. The set of Core and Voting ASes may be, but do not necessarily have to be the same ASes. Governance is implemented by a policy called the **Trust Root Configuration** (TRC), which is negotiated by the Voting ASes and which defines the locally scoped roots of trust used to validate bindings between names and public keys.
```
tzaeschke, line 178: Also mention authorative ASes here?

draft-dekater-scion-pki.md
```diff
@@ -161,48 +163,40 @@ SCION relies on three main components:
...
+Authentication in SCION is based on X.509 certificates that bind identifiers to public keys and carry digital signatures that are verified by roots of trust. SCION allows each ISD to define its own set of trust roots, along with the policy governing their use. Such scoping of trust roots within an ISD improves security as compromise of a private key associated with a trust root cannot be used to forge a certificate outside the ISD. An ISD's trust roots and policy are encoded in the TRC, which has a base and serial number, a list of public keys that serves as root of trust for various purposes, and a voting quorum governing the number of signatures required to update TRCs. The TRC serves as a way to bootstrap all authentication within SCION. Additionally, TRC versioning is used to efficiently revoke compromised roots of trust.
```
tzaeschke: `Such scoping of trust roots within an ISD improves security as compromise of a private key associated with a trust root cannot be used to forge a certificate outside the ISD.`

I'm not entirely sure, but certificates _are_ used outside of an ISD, e.g. when you want to communicate with an ISD?

draft-dekater-scion-pki.md
tzaeschke: LGTM

draft-dekater-scion-pki.md
tzaeschke: @nicorusti  If I understand section 1.4.1. correctly, a reset is only required for base TRC.

draft-dekater-scion-pki.md
tzaeschke: Maybe this was referring to "trust reset" and "update" as the two mechanisms?

draft-dekater-scion-pki.md
```diff
@@ -357,7 +351,7 @@ The RECOMMENDED maximum validity period of a sensitive voting certificate is 5 y
 | | - ID               - Description         | |
 | | - Validity         - No Trust Reset      | |
 | | - Grace Period     - Voting Quorum       | |
-| | - ...                                    | |
+| | - SerialNumber: 1  - ...                 | |
```
```suggestion
| | - SerialNumber     - ...                 | |
| | - ...                                    | |
```

draft-dekater-scion-pki.md
nicorusti: Saw your thumbs up, added in b4cc31a..26aadeb

draft-dekater-scion-pki.md
nicorusti: Let's briefly discuss tomorrow

draft-dekater-scion-pki.md
nicorusti: There are 3 mechanisms in total: 
- TRC updates and trust resets
- Short lived certificates

nicorusti: How about: 
```suggestion
The Control Plane PKI does not explicitly support certificate revocation. Instead it relies on the TRC update mechanism, on trust resets, and  on short-lived certificates. This approach constitutes an alternative to a revocation system for the following reasons:
```

draft-dekater-scion-pki.md
nicorusti: No hard feelings about this.. If we remove it from here, it appears in Figures 1 and 2 without any context or explanation

draft-dekater-scion-pki.md
nicorusti: This was introduced to address @oncilla's feedback in #149: 

> For TRC you should generally talk about serial and base numbers as to not confuse the reader. I think here, I would replace version number with serial number and gloss over the base number.

I didn't know where else to squeeze "1" in the figure :(

draft-dekater-scion-pki.md
```diff
@@ -357,7 +353,7 @@ The RECOMMENDED maximum validity period of a sensitive voting certificate is 5 y
...
```
```suggestion
| | - SerialNumber     - "1", e.g. for TRC 1 | |
| | - ...                                    | |
```
I think the `1` is in the header. Otherwise, how about this?

draft-dekater-scion-pki.md
nicorusti: the cause of the the reset is _any_ compromised TRC. So we have to say:

> If the ISD's ~base~ TRC has been compromised

 The result is a new _base_ TRC.

draft-dekater-scion-pki.md
nicorusti: Is it really relevant to the trust model? Or is it enough that we introduce them earlier in the terminology? 

IN the end, they are just a "distribution point" for TRC updates, so IMHO they don't really affect the trust model. No hard opinions on this

draft-dekater-scion-pki.md
nicorusti: I think we should reword to clarify that the "blast radius" of a compromised root or CA cert is limited to ASes within an ISD (and does not affect ASes in other ISDs)

```suggestion
Authentication in SCION is based on X.509 certificates that bind identifiers to public keys and carry digital signatures that are verified by roots of trust. SCION allows each ISD to define its own set of trust roots, along with the policy governing their use. Such scoping of trust roots within an ISD improves security as the impact of a compromised root or issuing CA certificate  is limited to within the affected ISD. An ISD's trust roots and policy are encoded in the TRC, which has a base and serial number, a list of public keys that serves as root of trust for various purposes, and a voting quorum governing the number of signatures required to update TRCs. The TRC serves as a way to bootstrap all authentication within SCION. Additionally, TRC versioning is used to efficiently revoke compromised roots of trust.
```

draft-dekater-scion-pki.md
nicorusti: This is what @tzaeschke said in the issue: 

> Remove last sentence with comparison to other PKIs, this sounds like unbacked advertisement.
 
 However we need this to orient readers that are searching for RPKI

draft-dekater-scion-pki.md
```suggestion
```

draft-dekater-scion-pki.md
nicorusti: discussed and agreed to not mention them here (since TRC updates are later on).

draft-dekater-scion-pki.md
```diff
@@ -161,48 +162,42 @@ SCION relies on three main components:
...
 
-While it is not necessary that all the ASes of the ISD trust each other, all ASes MUST trust the ISD's Core ASes, Authoritative ASes, Voting ASes, as well as its CA(s).
+While it is not necessary that all the ASes of the ISD trust each other, all ASes implicitly trust the ISD's Core ASes, Authoritative ASes, Voting ASes, as well as its CA(s).
 
 ### Updates and Trust Resets {#trust-reset}
 
-There are two types of TRC updates: regular and sensitive (see [](#update)). A **regular TRC update** is a periodic re-issuance of the TRC where the entities and policies listed in the TRC remain unchanged, whereas a **sensitive TRC update** is an update that modifies critical aspects of the TRC, such as the set of Core ASes. In both cases the base TRC remains unchanged.
+There are two types of TRC updates: regular and sensitive. The update type depends on which fields are changed (see [](#update)). In both cases the base TRC remains unchanged.
```
nicorusti, line 196:
```suggestion
There are two types of TRC updates: regular and sensitive. The update type depends on which fields are changed (see [](#update)). In both cases the base TRC remains unchanged.
Authoritative ASes announce these TRC updates (see [](#auth)). 
```

draft-dekater-scion-pki.md
nicorusti: So the wording about "Trust agility" was introduced in the very first version of the draft: 
https://github.com/scionassociation/scion-cppki_I-D/commit/206d8496c13d9eff8589612d905e0885067c8d5e back in Aug 2022 
This was in [-00](https://datatracker.ietf.org/doc/draft-dekater-scion-pki/00/)

This was also in the internal early draft ([link](https://github.com/scionassociation/standards/commit/afc57f7aeea272222e9805a3b6724beeb0e12f28))

nicorusti: - Do users/endpoints need provisioned TRCs to be able to do inter-ISD path loookups? If yes, they can really select roots of trust.. If not, they can't, but they can use path policies to select ISDs..

We need to find a good angle to this trust agility story. Either explain why, or skip on it. 
Till thinks it is a reference to EPIC. I cannot find it in the book or in the SCION docs.. I want to do more research to find where it comes form

draft-dekater-scion-pki.md
nicorusti: Other wording attempt:

```suggestion
Authentication in SCION is based on X.509 certificates that bind identifiers to public keys and carry digital signatures that are verified by roots of trust. SCION allows each ISD to define its own set of trust roots, along with the policy governing their use. This ISD-level scoping of trust roots enhances security by strictly limiting the blast radius of a compromise. If a root or issuing CA is compromised, it is restricted to issuing certificates solely for ASes residing within the affected ISD. An ISD's trust roots and policy are encoded in the TRC, which has a base and serial number, a list of public keys that serves as root of trust for various purposes, and a voting quorum governing the number of signatures required to update TRCs. The TRC serves as a way to bootstrap all authentication within SCION. Additionally, TRC versioning is used to efficiently revoke compromised roots of trust.
```

draft-dekater-scion-pki.md
```diff
@@ -161,48 +162,43 @@ SCION relies on three main components:
...
+Authentication in SCION is based on X.509 certificates that bind identifiers to public keys and carry digital signatures that are verified by roots of trust. SCION allows each ISD to define its own set of trust roots, along with the policy governing their use. Such scoping of trust roots within an ISD improves security as compromise of a private key associated with a trust root cannot be used to forge a certificate outside the ISD. An ISD's trust roots and policy are encoded in the TRC, which has a base and serial number, a list of public keys that serves as root of trust for various purposes, and a voting quorum governing the number of signatures required to update TRCs. The TRC serves as a way to bootstrap all authentication within SCION. Additionally, TRC versioning is used to efficiently revoke compromised roots of trust.
```
```suggestion
Authentication in SCION is based on X.509 certificates that bind identifiers to public keys and carry digital signatures that are verified by roots of trust. SCION allows each ISD to define its own set of trust roots, along with the policy governing their use. An ISD's TRC is used for signatures pertaining to information originating from that ISD, such as paths, but for nothing originating outside of the ISD. This ISD-level scoping of trust roots enhances security by strictly limiting effect of a compromise to data originating from the compromised AS. 
An ISD's trust roots and policy are encoded in the TRC, which has a base and serial number, a list of public keys that serves as root of trust for various purposes, and a voting quorum governing the number of signatures required to update TRCs. The TRC serves as a way to bootstrap all authentication within SCION. Additionally, TRC versioning is used to efficiently revoke compromised roots of trust.
```
Another attempt, trying to explain what is going on. 

And yes, we should avoid the term "blast radius", it is informal and not well defined. Also, in my opinion, the blast radius includes everyone that is affected, and that is everyone that has data (e.g. a path) that is related to the compromised AS.

draft-dekater-scion-pki.md
tzaeschke: -> I was refering to the last sentence of the previous paragraph: `The SCION trust model therefore, differs from the one provided by other PKI architectures.`.

draft-dekater-scion-pki.md
tzaeschke: TRCs are required for path validation. If an AS receives a path, it should validate the path with remote TRCs. So this trust agility essentially boils down to the ability to chose paths through ISD you trust and avoid paths through ISDs you don't trust.

draft-dekater-scion-pki.md
tzaeschke: LGTM

draft-dekater-scion-pki.md
knmeynell: Blast radius is a colloquialism that should't be used in a formal document (usually of course you're referring to military doctrine). "Strictly limiting the impact of a compromise" would be a better turn of phrase.

draft-dekater-scion-pki.md
nicorusti: Right, 1 is actually in the iD.SerialNumner field. I added it back on top in 84c2704936dd6f290185b3a71f34d04957e30460 , since there  is some space:

It looks good enough in both TXT and SVG
Screenshot 2026-05-28 at 09 57 14
Screenshot 2026-05-28 at 09 57 27
Screenshot 2026-05-28 at 09 58 04
Screenshot 2026-05-28 at 09 58 19

draft-dekater-scion-pki.md
nicorusti: Used Till's proposal in [a9533f7](https://github.com/scionassociation/scion-cppki_I-D/pull/146/commits/a9533f7884febc9d145460a657b74069a4dc52a0)

draft-dekater-scion-pki.md
nicorusti: I'd still like to add the reference to RPKI back in. How about: 

> To achieve trust agility, SCION avoids global PKIs such as the RPKI {{RFC8210}} where trust roots are provided by the Regional Internet Registries. Instead, 
in the CP-PKI, each ISD has its own trust root. Note that SCION does not provide IP prefix origin validation.

Old wording: 
> 
The need for trust agility also means that SCION does not by design provide IP prefix origin validation as provided by RPKI [RFC8210]. RPKI's trust model is currently reliant on the trust roots provided by the five Regional Internet Registries, and therefore outside of the governance of an ISD.

draft-dekater-scion-pki.md
nicorusti: Ok, so can we restore something closer to the older wording:

> This implies that users are free to choose an ISD they believe maintains an uncompromised set of trust roots. ISD members can also decide whether to trust other ISDs and thus transparently define trust relationships between parts of the network. The SCION trust model therefore, differs from the one provided by other PKI architectures.

draft-dekater-scion-pki.md
nicorusti: Re-added this wording above in 9ba7815

draft-dekater-scion-pki.md
nicorusti: Asked for more opinions on [Slack](https://scionproto.slack.com/archives/C8ADA9CEP/p1780045194860189)

draft-dekater-scion-pki.md
nicorusti: Changed in c3f43d3745cbf10e6f3e8f5c0ddae59adae70dc1

draft-dekater-scion-pki.md
```diff
@@ -163,48 +164,46 @@ SCION relies on three main components:
...
 
 ### Updates and Trust Resets {#trust-reset}
 
-There are two types of TRC updates: regular and sensitive (see [](#update)). A **regular TRC update** is a periodic re-issuance of the TRC where the entities and policies listed in the TRC remain unchanged, whereas a **sensitive TRC update** is an update that modifies critical aspects of the TRC, such as the set of Core ASes. In both cases the base TRC remains unchanged.
+There are two types of TRC updates: regular and sensitive. The update type depends on which fields are changed (see [](#update)). In both cases the base TRC remains unchanged.
+Authoritative ASes announce these TRC updates (see [](#auth)).
 
-If the ISD's TRC has been compromised, it is necessary for an ISD to re-establish the trust root. This is possible with a process called **trust reset** (if permitted by the ISD's trust policy). In this case, a new base TRC is created.
+If the ISD's TRC has been compromised, it is necessary for an ISD to re-establish the trust root. This is possible with a process called **trust reset** (if permitted by the ISD's trust policy) when a new base TRC is created.
```
```suggestion
If the ISD's TRC has been compromised, it is necessary for an ISD to re-establish the trust root. This is possible with a process called **trust reset** (if permitted by the ISD's trust policy). In this case, a new base TRC is created.
```
Why was this changed? If I understand correctly, `a new base TRC is created` when a trust reset is initiated, not the other way around?

```suggestion
If the ISD's TRC has been compromised, it is necessary for an ISD to re-establish the trust root. This is possible with a process called **trust reset** (if permitted by the ISD's trust policy). In this case, a new base TRC is created. See also [](#compromise-of-an-isd).
```
Section 6.1.1 explains when a trust reset is required and how it works. Maybe this part of section 6.1.1 should actually be moved here? Or to section 3.5?

nicorusti: It was some rewording by Kevin in https://github.com/scionassociation/scion-cppki_I-D/commit/c5d33d8426b2111243b3feda54dd48a0e38266cd

nicorusti: Since this affects 3 sections, let me deal with it in a later PR https://github.com/scionassociation/scion-cppki_I-D/issues/158

tzaeschke (approved):
LGTM

draft-dekater-scion-pki.md
```diff
@@ -163,48 +164,46 @@ SCION relies on three main components:
...
+An ISD's trust roots and policy are encoded in the TRC, which has a base and serial number, a list of public keys that serves as root of trust for various purposes, and a voting quorum governing the number of signatures required to update TRCs. The TRC serves as a way to bootstrap all authentication within SCION. Additionally, TRC versioning is used as an alternative to revocation in case of compromised roots of trust.
 
-The need for trust agility also means that SCION does not by design provide IP prefix origin validation as provided by RPKI {{RFC8210}}. RPKI's trust model is currently reliant on the trust roots provided by the five Regional Internet Registries, and therefore outside of the governance of an ISD.
+The TRC also provides *trust agility* - enabling users to select the trust roots used to initiate certificate validation. This implies that ASes are free to choose one or more ISDs they believe maintain an uncompromised set of trust roots. ASes can thus transparently define trust relationships between parts of the network, which differs from trust models provided by other established PKI architectures.
```
nicorusti:
```suggestion
The TRC also provides *trust agility* - enabling users to select the trust roots used to initiate certificate validation (e.g., to verify path segments). This implies that users are free to choose an ISD they believe maintains an uncompromised set of trust roots. Furthermore, AS member ASes of an ISD can decide whether to trust other ISDs and thus transparently define trust relationships between parts of the network. The SCION trust model therefore, differs from the one provided by other PKI architectures.
```

nicorusti: I'm restoring the earlier wording and deferring this to a dedicated issue

```suggestion
The TRC also provides *trust agility* - enabling users to select the trust roots used to initiate certificate validation. This implies that users are free to choose an ISD they believe maintains an uncompromised set of trust roots. ISD members can also decide whether to trust other ISDs and thus transparently define trust relationships between parts of the network. The SCION trust model therefore, differs from the one provided by other PKI architectures.
```

draft-dekater-scion-pki.md
nicorusti: Since we are not reaching an agreement about this, I'll merge this PR and opened https://github.com/scionassociation/scion-cppki_I-D/issues/159 to keep track of this point

nicorusti:
[Uploading Diff_ draft-dekater-scion-pki.txt - draft-dekater-scion-pki.txt.html…]()

Issue #149, opened 2026-05-13
Review Dominik


oncilla:
### 1.1. Terminology

> Voting ASes: Those ASes within an ISD that may sign TRC updates. The process of appending a signature to a new TRC is called "casting a vote".

It is technically possible that voters are not actual routable SCION ASes. That is actually also reflected later on by the recommendation that voters should have an ISD-AS number in their subject, but are not required to have it. If we want to be technically correct, we probably should use a different word than AS. For example, in the SSUN, we have a case where VSE is a voter (and we allocated an AS number), but for the time being they are not routable.

I am also fine with leaving it as is, just wanted to point this out in case it was not known/obvious.

### 1.3 

> , which has a version number, 

Version number here is imprecise. There is a field called version in the TRC payload which indicates the semantic version of the payload.
For TRC you should generally talk about serial and base numbers as to not confuse the reader. I think here, I would replace version number with serial number and gloss over the base number.

### 1.4

> While it is not necessary that all the ASes of the ISD trust each other, all ASes MUST trust the ISD's Core ASes, Authoritative ASes, Voting ASes, as well as its CA(s).

I wonder if we need to distinguish the level of trust that we need to have in these entities. For the CP-PKI, we do not really need to trust Core or Autoritative ASes. CAs and Voters need to be trusted though.

### 2.6.4

> In issuing CA certificates, the presence of this attribute ensures that the Control Plane can identify from which AS to retrieve the intermediate CA certificate, thereby avoiding circular dependencies

That sentence does not really make sense. We never have to retrieve the intermediate CA certificate. The AS certificate is always bundled together with the intermediate CA certificate as a certificate chain.

### 2.7.1 

> authorityCertIssuer & authorityCertSerialNumber: MUST NOT be included. Implementations MUST return an error if either is present.

As also mentioned elsewhere, currently this is not enforced by our implementation. I am fine however to add this, as we are not setting these fields in our current implementations.

### 3

> The TRC is a signed collection of [[X.509](https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.html#X.509)] v3 certificates. Additionally, the TRC contains ISD-specific policies encoded in CMS signed-data ([[RFC5652](https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.html#RFC5652)] section 5).

This makes it sound like the certificates are not part of the CMS signed-data. I think it could be rephrased to be clearer.

### 3.1

[...]

PR #151, opened 2026-05-13
chapter 2 review

### LINKED ISSUE #143: Review Til Chpt 2
## 2.1 

- "proof of ownership of the private key" who owns these private keys? If they are owned by "TRC", why have multiple?
- " This facilitates the bootstrapping of trust within an ISD". This process is not clear to me. Maybe give step by step instructions of who does what? Or reference the chapter if it is described later.

- " 5 years." -> Why not 1 year? 5 years means there is no operational routine and it is a first for many operators each time it happens....

## 2.2.
- "must first be extracted from one or more active TRCs" Why do we need to consider "more" TRCs? Only during the grace period, right?

## 2.3
- "3 days" -> What about the grace period? Paths can be valid for a long time.

## 2.4
- just for clarity, add "There is one private key per voting AS (and per sensitive/regular)". I.e. not 0 keys and not more than 1 key, right?

- In this chapter, maybe explain the rationale for different certificates for regular/sensitive?  Table 7 explains the difference and 3.5.6 gives a motivation. Maybe just reference 3.5 here?

- 2.4.1 and 2.4.2 should be merged, they have identical text except for regular vs sensitive.

## 2.5
- table 2: 
  - (1) Explain why it makes sense to include multiple certificates
  - (3) Why 8 days?
  - ... What is the overlap for other certificates?
 - What is "TRC 1"? Is it a name or number stored in the TRC? Or is it just a virtual name used here?

## 2.6
- `whilst` -> `while`
- Simplify 1st sentence from 
  ``` 
  Whilst the certificates used in the Control Plane PKI are X.509 v3 certificates, this specification is more restrictive.
   ``` 
   to
  ```
  Control Plane PKI certificates are X.509 v3 certificate with additional constraints applied.
  ```  

### 2.6.1: 
- Clarify that this is the X.509 version and not the TRC version. 
- Clarify that this refers to X.509 extensions (not SCION extensions; this is the first time extensions are mentioned)
- Assuming this is not the TRC version, where actually is the TRC version stored? Or is it implicit from the issuing date (if there is such a thing)?

### 2.6.2: 
- Digits? Hex/decimal? 

### 2.6.4: 
- Also, maybe the issuer name MUST be unique? (at least within the scope of an ISD?

#### 2.6.4.1: 
- What is the format of "issuer" and "subject", how is "id-at-ia" included? Separator? Is appended or prepended? Sorry if there is a standard way of doing this and this question makes no sense.

### 2.6.5
- "SCION implementations SHOULD adopt these values." Change
  "implementations" -> "deployments"? I assume this is configurable and not hardcoded?

### 2.6.6
- Is it worth being more specific than "same constraints as issuer"? There is only one additional constraint in "issuer" in addition to not being empty. Rephrase to `... it MUST NOT be empty and in addition include the "id-at-ia" field.`?

### 2.6.8/2.6.9 
- Maybe out of scope, but a brief mention _why_ they should not be used could be interesting.

### 2.7.1: 
- "non-critical" I assume "critical refers to something in X.509, maybe add "... non-critical in the sense of RFC0000"?
- Last sentence: So it must be accepted if extension is missing but the certificate is self-signed? What is the rationale?

### 2.7.2: 
 - "It can be used, for example, by control plane messages to identify which certificate to use for verification."
 - ` It can be used, for example, by control plane messages to identify which certificate to use for verification.` I think this implies that implementations MUST be able to use this to identify which certificate to use? Is this specified anywhere?
- Last sentence: So it must be accepted if extension is missing but the certificate is self-signed? Rationale?

### 2.7.3:
- `digitalSignature: The public key can be used to verify the digital signature of a control plane payload.` ->
  If it can be used, this implies implementations MUST support this?

- `keyCertSign: The public key can be used to verify the CA signature on a control plane AS certificate.` ->
  If it can be used, this implies implementations MUST support this?

- `Other attributes are not used.` -> `Other attributes MUST NOT be used`?

- `...  the relying party MUST be able to trace back the private key used to sign the certificate`
  How is this MUST actionable? "MUST be able to do something" seems a strange use of "MUST". Is this a requirement on implementations?

### 2.7.4 
- "Note":  `Root` -> `root`
- `... implementations follow what --is-- described in this document` 
- `the use of extKeyUsage should be revised in future protocol iterations`
  This is a strange statement. Maybe explain why it should be revised? Why is that mentioned here?

### 2.7.5
- The reference to X.509 points to a "superseded" document

### 3.2.6 
- Maybe add some suggestions on the value of this field? E.g. should be (voting members - 2) two allow for removing members? Or if a member loses its certificates (and two events overlap)?

nicorusti:
Resolves #143 
Contributes to #149 

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.txt&url_2=https://scionassociation.github.io/scion-cppki_I-D/til-chapter2/draft-dekater-scion-pki.txt)


draft-dekater-scion-pki.md
```diff
@@ -300,21 +300,13 @@ The RECOMMENDED maximum validity period of a CP AS certificate is 3 days.
 
 ## Voting Certificates {#cp-voting-cert}
 
-There are two types of voting certificates: the (1) regular voting certificates and the (2) sensitive voting certificates. They contain the public keys associated with the private keys that may cast votes in the TRC update process.
+There are two types of voting certificates: regular voting certificates and sensitive voting certificates. They contain the public keys associated with the private keys that may cast votes in the TRC update process.
 
-Regular and sensitive voting certificates are used to verify regular and sensitive TRC updates respectively, and are embedded in the TRC.
+Regular and sensitive voting certificates are used to verify regular and sensitive TRC updates respectively, and are embedded in the TRC. The distinction between segular and sensitive updates is described in [](#update).
```
knmeynell: *regular* and sensitive updates

knmeynell (changes requested):

draft-dekater-scion-pki.md
```suggestion
Regular and sensitive voting certificates are used to verify regular and sensitive TRC updates respectively, and are embedded in the TRC. The distinction between regular and sensitive updates is described in [](#update).
```

nicorusti:
Comments are addressed until 2.4 so far

draft-dekater-scion-pki.md
```suggestion
If a relying party uses the certificate’s public key to verify the signature of a control plane payload (`digitalSignature` attribute), the relying party is able to trace back the private key used to sign the certificate. This is done by referencing the ISD-AS and the subject key identifier (via the `subjectKeyIdentifier` extension). For more information about the `subjectKeyIdentifier` extension (see [](#subject-key-id-ext)).
```

tzaeschke: See https://github.com/scionassociation/scion-cppki_I-D/issues/143#issuecomment-4442258914

tzaeschke (commented):
LGTM except my comment on 2.7.3

draft-dekater-scion-pki.md
nicorusti: Uups I forgot to push, this was in [a439233](https://github.com/scionassociation/scion-cppki_I-D/pull/151/commits/a4392336daa2dbdf5713b7f410992993e001eaf7)

nicorusti:
[Diff_ draft-dekater-scion-pki.txt - draft-dekater-scion-pki.txt.html](https://github.com/user-attachments/files/28544083/Diff_.draft-dekater-scion-pki.txt.-.draft-dekater-scion-pki.txt.html)

PR #152, opened 2026-05-14
Til-Chapter-3

### LINKED ISSUE #144: Review Til Chpt 3
## 3.1
- Under `A TRC can have the following states:`, add 
  `- invalid: The TRC is neither valid not active`
   
### 3.2.2
- `baseNumber`: This starting point is _always_ the currently valid base TRC, no special case for the initial TRC. Replace
   "This starting point is either the ISD's initial TRC or the currently valid base TRC" 
   with 
   "This starting point is the currently valid base TRC"  

### 3.2.3

Font of last "GeneralizedTime" is inconsistent

### 3.2.4
- Remove the "or" after the second bullet, I don't think "or" is used with ";".
- Suggested value for grace period?

### 3.2.7
- Recommend upper limit for quorum? Using the maximum is risky because certificates may get lost, or a rogue AS may refuse to vote...?

### 3.2.9
- Suggested number of authorative ASes? (compared to core/voting ASes)
- Should they overlap with voting ASes? Why not the same?

### 3.2.10
- Should there be a designated separate for non-English text?
- Should there be a designator for the (non-English) language?

### 3.2.11
- "The issuer / serialNumber pair for each certificate MUST be unique."
  Is it even possible to create different certificates with the same issuer + serialNumber? If not, isn't this condition implied in the previous point?

knmeynell:
Resolves #144 
Contributes to #155

- Addressed 3.1, 3.2.2, 3.2.3 & 3.2.4
- Do we have suggestions for 3.2.7, 3.2.9, 3.2.10 & 3.2.11?

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.txt&url_2=https://scionassociation.github.io/scion-cppki_I-D/Til-Chapter-3/draft-dekater-scion-pki.txt)


draft-dekater-scion-pki.md
```diff
@@ -678,7 +679,7 @@ The `validity` field defines the TRC validity period. The `notBefore` and `notAf
 An active TRC is a valid TRC that can be used for verifying certificate signatures. The time period during which a TRC is active can be shorter than the time period during which the TRC is valid. For more information, see [](#trc-states).
 
 The `validity` field consists of a sequence of a `notBefore` and a `notAfter` date, both encoded as `GeneralizedTime`.
-All TRCs MUST have a well-defined expiration date. SCION implementations MUST NOT create TRCs that use GeneralizedTime value "99991231235959Z", and verifiers MUST reject such a TRC.
+All TRCs MUST have a well-defined expiration date. SCION implementations MUST NOT create TRCs that use GeneralizedTime` value "99991231235959Z", and verifiers MUST reject such a TRC.
```
oncilla: That looks off

oncilla (changes requested):

draft-dekater-scion-pki.md
```suggestion
All TRCs MUST have a well-defined expiration date. SCION implementations MUST NOT create TRCs that use `GeneralizedTime` value "99991231235959Z", and verifiers MUST reject such a TRC.
```

draft-dekater-scion-pki.md
```diff
@@ -631,6 +632,7 @@ A TRC can have the following states:
 
 - Valid: The validity period of a TRC is defined in the TRC itself, in the `validity` field (see [](#validity-trc)). A TRC is considered valid if the current time falls within its validity period.
 - Active: An active TRC is a valid TRC that can be used for verifying certificate signatures. This is either the latest TRC or the predecessor TRC if it is still in its grace period (as defined in the `gracePeriod` field of the new TRC, see [](#grace)). No more than two TRCs can be active at the same time for any ISD.
+- Invalid: The TRC is neither valid nor active.
```
nicorusti:
```suggestion
- Invalid: The TRC is considered invalid if the current time falls outside its validity period.
```

Id does not make sense to mention "active" since the active time window is a subset of the valid time window

nicorusti:
@tzaeschke ready for you to review

draft-dekater-scion-pki.md
```diff
@@ -724,7 +726,7 @@ The `votes` sequence MUST be present to prevent the stripping of voting signatur
 The `votingQuorum` field defines the number of necessary votes on a successor TRC to make it verifiable.
 
 A voting quorum greater than one will prevent a single entity from creating a malicious TRC update.
-
+A voting quorum lower than the number of Voting ASes ensures that votes can be cast even if some of the voters are unavailable.
```
nicorusti, line 729: @tzaeschke like it ?

tzaeschke: Yes, I like it.

draft-dekater-scion-pki.md
tzaeschke, line 743: What is missing here is a recommendation on:
- How many authorative ASes should have?
- How do we chose them (is it okay if they are voting ASes?)
- Do authorative ASes require additional infrastructure?

And some explanation what their purpose is:
- Why do we need them? They are not part of the trust model... 
- If there is no purpose beyond just giving the concept a name, maybe state just that. As it is, the reader is left wondering whether there is importance or not.

tzaeschke: Maybe from the AS perspective:
- What are the advantages/disadvantages for an AS of being authorative?

tzaeschke, line 949: This section (Section 3) explains concepts such as "trust updates" and "the signing ceremony", but it does not describe "trust reset", which is a very related concept and basically the middle ground..?.
I think this is missing.
Section 1.4.2 gives an overview over trust reset. Section 6.1.1 goes into details.

tzaeschke, line 943:
```suggestion
## Trust Reset

If the ISD's TRC has been compromised, it is necessary for an ISD to re-establish the trust root. This is possible with a process called trust reset (if permitted by the ISD's trust policy). In this case, a new base TRC is created.

If a TRC is compromised, it means that the root keys or voting keys are compromised. If the number of compromised keys is smaller than the voting quorum, the TRC can be updated as described in [Section 3.5](https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.html#update). 

A trust reset is only required in the case the number of compromised keys at the same time is greater or equal than the TRC's quorum (see [Section 3.2.7](https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.html#quorum)), and a invalid update has been produced and distributed in the network.

```
This section (Section 3) explains concepts such as "trust updates" and "the signing ceremony", but it does not describe "trust reset", which is a very related concept and basically the middle ground..?.
I think this is missing.
Section 1.4.2 gives an overview over trust reset. Section 6.1.1 goes into details.

The text above is an amalgamation of 1.4.1 and 6.1.1. Section links need to be fixed.

tzaeschke: This comment can be resolved, I moved it to the suggestion above.

nicorusti: Postponing to separate issue since I have to touch 3 sections to have proper cross references
https://github.com/scionassociation/scion-cppki_I-D/issues/158 
I copied your proposal over

nicorusti: > Why do we need them? They are not part of the trust model...
 
 This is already explained in the terminology
 >    *Authoritative AS*: Authoritative ASes are those ASes in an ISD that
   always have the latest TRCs of the ISD.  As a consequence,
   authoritative ASes also start the announcement of a TRC update. 
 
 And also in this paragraph you quote:
 
 > Authoritative ASes are those ASes in an ISD that always possess the latest TRCs for the ISD and initiate TRC update announcements.
 
For the other points, I am not sure if this fits best 
3.2.9.  authoritativeASes
4.1.2.  TRC Update Discovery (we should mention the role of authoritative ASes here!)
5.  Deployment Considerations (maybe we could say something about the availability requirements)

Since this affects 3 sections tracking this in https://github.com/scionassociation/scion-cppki_I-D/issues/155

tzaeschke (approved):

nicorusti:
[Diff_ draft-dekater-scion-pki.txt - draft-dekater-scion-pki.txt.html](https://github.com/user-attachments/files/28396321/Diff_.draft-dekater-scion-pki.txt.-.draft-dekater-scion-pki.txt.html)

PR #153, opened 2026-05-22
ASN1 - lower max size for sequences

nicorusti:
To avoid infinite size TRCs

Hi @oncilla , @tzaeschke is suggesting that we introduce stricter upper bounds for sequence sizes, to prevent having unbounded bloated TRCs that could be a DoS attack vector. 

We currently have a potentially unlimited size:
- 4095 certificates (16MB assuming 4KB/certificate)
- For each description field, max 8192 
- MAX localizedDescriptions (ubounded sequence - infinite)
- coreASes - unlimited
- authoritativeASes - unlimited
- ASN - unlimited
- language / descriptionLanguage unlimited

Shall we reduce limits a bit?  Something like:
- 1023 total certificates (4MB) 
- 128 localizedDescriptions (sequence - 1MB)
- For each description field, max 8192 (same as before? - I don't think it makes a big difference)
- ASN - 16 bytes
This way a TRC will be around 5MB, except if some certs are super bloated..

Remaining problem: 
- coreASes & authoritativeASes  are still unlimited - the SCION limit is 2**48, which would result in a still gigantic TRC.. Do we want to introduce a cap here? Why, and to what? there are no other hard limits in the spec AFAIK
@oncilla 

TODO: 
- [x] copy changes into appendix before merging

### DIFF
--- a/asn1/trc.asn1
+++ b/asn1/trc.asn1
@@ -29,13 +29,13 @@ TRCPayload ::= SEQUENCE {
     validity              TRCValidity,
     gracePeriod           INTEGER,
     noTrustReset          BOOLEAN,
-    votes                 SEQUENCE SIZE (0..2047) OF INTEGER (0..4095),
-    votingQuorum          INTEGER (1..2047),
+    votes                 SEQUENCE SIZE (0..511) OF INTEGER (0..1023),
+    votingQuorum          INTEGER (1..1023),
     coreASes              SEQUENCE OF ASN,
     authoritativeASes     SEQUENCE OF ASN,
     description           UTF8String (SIZE (1..8192)) OPTIONAL,
-    certificates          SEQUENCE SIZE (1..4095) OF Certificate,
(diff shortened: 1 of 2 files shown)


nicorusti:
These in the PR are proposed limits, I would like to get input from Till and @oncilla
If we need, we can also ask Russ, an expert ASN.1 reviewer

tzaeschke:
Is there a possibility to impose a global limit, e.g. TRCs larger than 16MB are rejected?
This may be simpler than adjusting individual limits...

nicorusti:
@oncilla @tzaeschke I updated the PR to include: 
- Mention that TRC size SHOULD NOT exceed 4 MB
- Introduce only sequence/field limits only where it absolutely makes sense (language tag, ASN string, limit to 1024 languages)

Does that work for you ?

tzaeschke (approved):
LGTM

oncilla (approved):

PR #156, opened 2026-05-26
Review chapters 4-6 Till & Dominik

### LINKED ISSUE #150: Review Til Chpt 4, 5, 6 and Appendix C
### 4.1.2

- `Relying parties MUST have at least one valid TRC available.`
  Maybe specify what relying parties are. I assume it is control services of every AS in the ISD (== AS operators/infrastructure)?
  I assume it is not border routers and not endhosts.

- The 1st MUST does not give any instructions on what to do or how to achieve it, usually I expect MUST before a description of an  action, not before a description of a state. I suggest moving it to the 3rd point, see below.
- The 2nd MUST overlaps with the first, I suggest moving it to the 3rd bullet.
- How to proceed if any of these MUSTs fail (e.g. due to network separation)?

- Beaconing + path discovery: This section is about _ discovering_  TRC updates, but it contains normative language about CP traffic creation. Moreover these MUSTs appear to duplicate the MUSTs in section 4.2.1. Maybe remove the normative language here?

- There is an empty line between 2nd and 3rd point.

- Active discovery: `Any TRC can be obtained at any time from the sender of the information it secures; `
  This sentence is unclear to me, what is "it"?
  Also, why `any` TRC? Is should probably only get the latest one?
  Maybe rephrase to: `The latest TRC can be actively queried from any entity (control service) that has it.`

- `Note`: What is the point of this note, why is this relevant? Also, it seems incorrect, also the third mechanism involves active communication.
  Moreover, the first two mechanisms require the third one, correct?

Suggestion: 
  - Before the bullet list, insert: `New TRCs can be obtained from any AS that has them. The necessary query and response is described in [I-D.dekater-scion-controlplane], section "Distribution of Cryptographic Material"`.
  - Change third bullet to `Any relying party can actively poll other parties for new TRC.`
  - Add a separate list below the bullet list:  
    ```
    When to poll:
    - A relying party MUST poll if the current TRC is about to expire.
    - A relying party SHOULD poll if path lookup or beaconing traffic is sparse
    - A relying party MAY poll at any time.
    ``` 
  - Remove `Note: ...`

## 4.2
- `parties MUST be able`  What is the implication here? What must an implementer do to achieve this? Or does this follow implicitly from following from other sections of this document (including the following sections)? In that case maybe simply replace this `MUST be` with `are`?

### 4.2.1
- remove parenthesis around `(valid)` or simplify/rephrase to `An AS signs control plane messages with the private key that corresponds to its own certificate.`

### 4.2.2
- remove reference to SCION: `For efficiency, SCION distributes these certificate chains separately from the signed messages.`
  -> `For efficiency, these certificate chains are distributed separately from the signed messages.`

- 2nd point: Why is this a MUST? If this is security critical, why is this information optional for the sender? 
- 2nd point: How does one know whether this is the `latest TRC`? Do they have to perform a TRC query to check?
- 2nd point: Combine two MUST into one: `... the relying party MUST ensure that they have this latest TRC.`

- 3rd point: Remove 1st MUST, it seems unnecessary ->  `the relying party selects the certificate chain that will be used to verify the message`
- 3rd point: Remove the three sub-MUSTs, they are duplicates, the first MUST implies the other 3. Maybe make it similar to 4th point.
- 3rd point: Is it worth adding something about the `current time`? I.e. that "coarse synchronization` is sufficient (reference CP draft 7.3? Maybe a reference to a section that discusses time tolerance?

- 5th point: What happens if the verification or the check fails?

- It seems a bit inconsistent that the failure of (4.) is discussed in a separate point (5.) while the failure of (3.) is discussed in a separate paragraph afterwards.

- `Important`: 
  - Remove normative SHOULD, it is not actionable. Also an "implication" usually means that something "is" or "is not", rather than "should".

  - I'm not sure what is said here. Why SHOULD they, or rather, may not be, verifiable? As long as I have a certificate, I can verify them (verification may fail because the certificate is outdated)?
    Is this meant to say: "An implication of the above procedure is that path segments SHOULD be _valid_ at time of use."?
    How is this an implication of the procedure, the procedure says nothing about storing the path and using it later...
    Maybe remove the first sentence and shorten the whole paragraph to "Important: It is not enough to rely on path segments being verified on insert since TRC updates that change the root key can invalidate a certificate chain."?

  - Does "It is not enough" mean that something MUST be done? -> Insert MUST

## 4.3
- Does "REQUIRED" need to be normative? This is just a listing of steps, they technically require each other implicitly.
- Does "MAY" need to be normative?
  

## 5.1
- `PKI operators need to ensure that the CAs maintain accurate time.`
  In CP-draft, it is not called "accurate" but "coarse", maybe write `PKI operators need to (MUST?) maintain coarse time synchronization as discussed in [I-D.dekater-scion-controlplane], section "Attacks on Time Sources"`.
- Also, maybe replace `Attacks on Time Sources` with `7.3. Effects of Clock Inaccuracy`?

- `This document describes a typical TRC signing ceremony in Appendix C, but further processes are out-of-scope.` 
  `This document` is misleading, the previous sentence mentions a document, but a different one. How about: `The documentation of a typical TRC signing ceremony is discussed in Appendix C.`

# 6
- `This section discussed implication of such trust architecture, covering inter-AS security considerations.`
   Typo: past tense and singular: `discusseS implicationS`

## 6.1
- `In SCION there is no central authority that could "switch off" an ISD as each`
  add `In SCION there is no central authority that could "switch off" an ISD (except the ISD itself) as each ...`

- 1st point: Maybe simplify to  `... To maliciously perform a TRC update, an attacker would need to compromise enough voting keys for a voting quorum. The higher the quorum, the _harder_ a malicious update _becomes_.`

## 6.2
- `On the other hand,` -> `However,`  .  On the other hand usually comes after "on one hand"

- 1st point: Is this a SHOULD?
- This is not obvious to me. Every path segment is signed with a specific CA, right? So does that mean every AS should sign its PCBs with different CAs round-robin? If so, maybe that is worth making explicit? 
  Generally, can we be more specific in how one "relies" on multiple CAs?

# Appendix C

- In general, the signing procedure document will not be an RFC. Should there be any normative language here?
  Also, it says the procedure is RECOMMENDED, implying that all the following MUSTs are recommended, so the MUSTs are all optional .... ?
  Maybe we should remove all normative language from this section, including the "RECOMMENDED"?

## C.1
- The SHOULD and the previous RECOMMENDED seem redundant.
- `Voting ASes` is spelled with upper case here (several occurrences) but lower case elsewhere. Change to lower case?
- 2nd point `They are in possession of a device with the private keys of their respective certificates in the TRC.`
  Device? Maybe shorten to: `They are in possession of the private keys of their respective certificates in the TRC.`

- remove ` and issues encountered during the Signing Ceremony may be assumed to be caused by honest mistakes and not by malicious intent.`
- The sentence `Hash comparison checks are included to counter mistakes and so that every participant can ensure they are operating on the same data, and the private keys of each participant never leave their machine.` seems wrong: 
  - "Hash comparison checks are included" -> where? by whom? should they be part of a protocol? What should be hashed? The private keys? Why?
  - How does hash comparison _ensure_ that a private keys never leave a machine?
  - Mentioning of "avoiding mistakes" seems unnecessary

## C.2
- The initial sentence seems off:
  - Do we really need to specify that participants need to agree on a meeting location?
    If anything we could suggest to use a _physical_ _in-person_ meeting, but the location is irrelevant. 
  - It sounds like, physical location and devices are not a MUST, so why even mention it?
  - `MUST decide the ISD policy as follows;` makes it sound like the list gives instruction on how to decide on a policy. That is not the case.
    Maybe: `An ISD policy MUST include the following items:`?

- 1st point: change to `ISD number - for public ISDs obtained from the SCION registry, see Section 3.2.2;`

- `Voting AS` is upper case...

- `Each representative of a Voting AS MUST also create the following before the ceremony:`
  Why MUST _before_ the ceremony? Also, maybe remove the MUST, it seems implicit/obvious for providing voting keys.

- `In addition,.. ` all the MUSTs here seem obvious/implicit, I think they can be removed.

- `The location should provide electricity and power sockets for each participant, and should provide a monitor or projector that allows the Ceremony Administrator to display proceedings.`
  I'm not sure, does this really belong here?
 
- `The Ceremony Administrator MUST provide or be provided with a device to exchange data between the ceremony participants.`
  Again, these seems obvious to me. I would either remove the sentence or remove the MUST, it does not even specify who is actionable. 

- `SHOULD`: So after all these MUSTs, the security of the devices is optional?

### C.3.1
- Simplify sentence: `All parties share the certificates that must be part of the TRC with the Ceremony Administrator.` to 
  `All certificates that must be part of the TRC must be shared with the Ceremony Administrator.`

- `The Ceremony Administrator MUST then` this MUST seems implicit (= a precondition) for the following MUST. Remove it?
- Does it need to be SHA-512? What is wrong with SHA-3 or others?
  Maybe add `SHA-512 (or any equivalent or better algorithm)`?
  Then again, this document is only a recommended example, so maybe we can leave it as is.

### C.3.2
- 1st sentence: remove words `the` before Section 3.2 and `completed`?

- "DER encoding", provide reference?

- `by computing this on their` -> `by computing IT on their`
- `checking it matches` -> `checking THAT it matches`

### C.3.4
- `The Signing Ceremony is completed once when every voting representative` -> remove `once` or `when`

nicorusti:
Resolves #150 
Contributes to #149

### DIFF
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -960,21 +960,15 @@ Base TRCs are trust anchors and thus axiomatically trusted. All ASes within an I
 
 All non-base TRCs of an ISD are updates of the ISD's base TRC(s). The TRC update chain consists of regular and sensitive TRC updates. The specifications and rules that apply to updating a TRC are described in [](#update).
 
-Relying parties MUST have at least one valid TRC available. Relying parties MUST discover TRC updates within the grace period defined in the updated TRC, and SHOULD discover TRC updates in a matter of minutes to hours. Additionally, any entity sending information that is secured by the Control Plane PKI MUST be able to provide all the necessary trust material to verify said information.
+Relying parties such as an AS Control Service require at least one valid TRC available and should therefore discover TRC updates within the grace period defined in the updated TRC. Additionally, any entity sending information that is secured by the Control Plane PKI MUST be able to provide all the necessary trust material to verify said information, ensuring that relying parties can discover TRC updates in a matter of minutes to hours.
 
 SCION provides the following mechanisms for discovering TRC updates and fulfilling the above requirement:
 
-- *Beaconing Process*<br>
-The TRC version is announced in the beaconing process. Each AS MUST announce what it considers to be the latest TRC, and MUST include the hash value of the TRC contents to facilitate the discovery of discrepancies. Therefore, relying parties that are part of the beaconing process discover TRC updates passively, i.e. a Core AS notices TRC updates for remote ISDs that are on the beaconing path. A non-core AS only notices TRC updates for the local ISD through the beaconing process. The creation of a new TRC SHOULD trigger the generation of new control plane messages, as the propagation of control plane messages will help other ASes rapidly discover the new TRC.
+- *Beaconing Process*: The TRC version is announced during the beaconing process (see {{I-D.dekater-scion-controlplane}} section "AS Entry Signed Header"). Each AS MUST announce the latest TRC's base and serial number known to it. Consequently, relying parties involved in the beaconing process discover TRC updates passively, i.e. a Core AS notices TRC updates for remote ISDs that are on the beaconing path. A non-core AS only notices TRC updates for the local ISD through the beaconing process.
 
-- *Path Lookup*<br>
-In every path segment, all ASes MUST reference the latest TRC of their ISD. Therefore, when resolving paths, every relying party will notice TRC updates, even remote ones.<br>
+- *Path Lookup*: In every path segment, all ASes MUST reference the latest TRC of their ISD. Consequently, relying parties will detect TRC updates, including those from remote ISDs, during path lookups.
 
-- *Active Discovery*<br>
-Any TRC can be obtained at any time from the sender of the information it secures; either in a specific version or in its latest available version. The necessary query and response is described in {{I-D.dekater-scion-controlplane}}, section "Distribution of Cryptographic Material".
-
-**Note:** The first two mechanisms above only work when there is active communication between the relying party and the ISD in question.
-
+- *Active Discovery*: A relying party can actively request any TRC —either a specific version or the latest available version— from the sender of the secured information at any time. The necessary query and response is described in {{I-D.dekater-scion-controlplane}}, section "Distribution of Cryptographic Material".
 
 
 ## Signing and Verifying Control Plane Messages {#signing-verifying-cp-messages}
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -979,7 +979,7 @@ The following sections specify the requirements that apply to the signing and ve
 
 ### Signing a Control Plane Message
 
-An AS signs control plane messages with the private key that corresponds to the (valid) AS' certificate.
+An AS signs control plane messages with the private key that corresponds to its own valid certificate.
 
 The AS MUST attach the following information as signature metadata to ensure that a relying party can identify which certificate to use to verify the signed message:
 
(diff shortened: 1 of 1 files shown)


nicorusti:
Kevin did more work in #157 . CLosing this PR, ported my two commits to the other one

PR #163, opened 2026-06-03
cert validity: clarify that renewal should happen before hop field ex…

### LINKED ISSUE #148: Double check cert max validity
From #143: 

> 2.3
"3 days" -> What about the grace period? Paths can be valid for a long time. 

There is no grace period for AS certificates. Two certificates can be valid at the same time, and during verification the relying party selects the right cert based on the  subject key identifier  (see https://www.ietf.org/archive/id/draft-dekater-scion-pki-12.html#name-verifying-a-control-plane-m ). We could add some wording saying that they are renewed before the old one expires, with a cross reference to that section. 

On path segments validity, in -controlplane, we say: 
```
2.2.2.6.  AS Entry Signature

   Each AS entry MUST be signed with the AS certificate's private key
   K_i.  The certificate MUST have a validity period that is longer than
   the Hop Field absolute expiration time (described in
   Section 2.2.2.5). 
```
A hop field max validity is 24h, so with 3 days you should have enough time to overlap and keep beaconing.. However, we should think if we want to extend it.

nicorusti:
Resolves #148

### DIFF
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -299,6 +299,7 @@ SCION ASes sign control plane messages, such as Path Construction Beacons, with 
 In X.509 terms, control plane AS certificates are end entity certificates. That is, they cannot be used to verify other certificates.
 
 The RECOMMENDED maximum validity period of a CP AS certificate is 3 days.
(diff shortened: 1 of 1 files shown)


tzaeschke (approved):
LGTM

PR #166, opened 2026-06-03
Trust Reset - consolidate concepts in dedicated section

### LINKED ISSUE #158: Clarify and consolidate trust reset process
From @tzaeschke: > I also noticed another problem (I also added a note the the PR of section 3: There is no real description of "trust reset". The closest thing is section 6.1.1, but that feels somewhat hidden. For comparison, there are section about "trust update" and "signing ceremony".

Indeed, trust resets are spread across various sections: 

```
1.4.1.  Updates and Trust Resets

   If the ISD's TRC has been compromised, it is necessary for an ISD to
   re-establish the trust root.  This is possible with a process called
   *trust reset* (if permitted by the ISD's trust policy).  In this
   case, a new base TRC is created.
```

```
3.2.5.  noTrustReset

   The noTrustReset Boolean specifies whether a trust reset is forbidden
   by the ISD.  Within a TRC update chain, this value MUST NOT be
   changed by a regular or sensitive update.  However, it is possible to
   change the noTrustReset value in the event of a trust reset where a
   new base TRC is created.

   The noTrustReset field is OPTIONAL and defaults to FALSE.

   Note that once the noTrustReset Boolean is set to TRUE and a trust
   reset is disallowed, this cannot be reversed.  Therefore, ISDs SHOULD
   always set this value to FALSE, unless they have sufficiently
   assessed the risks and implications of making a trust reset
   impossible.

   Note that a trust reset represents a special use case where a new
   base TRC is created.  It therefore differs from a TRC update (regular
   or sensitive) as the signatures in the new base TRC cannot be
   verified with the certificates contained in the predecessor TRC.
   Instead, a trust reset base TRC must be axiomatically trusted,
   similarly to how the initial TRC is trusted.

```

```
6.1.1.  Recovery from Compromise

   This section deals with possible recovery from the compromises
   discussed in the previous paragraph.  As described in Section 1.4.2,
   there is no revocation in the Control Plane PKI.

   *  At TRC level: If any of the root keys or voting keys contained in
      the TRC are compromised, the TRC MUST be updated as described in
      Section 3.5.  A trust reset is only required in the case the
      number of compromised keys at the same time is greater or equal
      than the TRC's quorum (see Section 3.2.7), and a invalid update
      has been produced and distributed in the network.
```

Let's check if we can consolidate and cross reference it

nicorusti:
Resolves #158 

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.txt&url_2=https://scionassociation.github.io/scion-cppki_I-D/trust_reset/draft-dekater-scion-pki.txt)

### DIFF
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -928,6 +928,11 @@ To verify a TRC update, the relying party MUST perform the following checks:
 
 If one or more of the above checks gives a negative result, the updated TRC SHOULD be rejected.
 
+## Trust Reset {#trust-reset-description}
+
+If the ISD's TRC has been compromised, it is necessary for an ISD to re-establish the trust root. This is possible with a process called trust reset (if permitted by the ISD's trust policy). In this case, a new base TRC is created.
+If a TRC is compromised, it means that the root keys or voting keys are compromised. If the number of compromised keys is smaller than the voting quorum, a TRC update is sufficient [](#update).
+A trust reset is only required in the case the number of compromised keys at the same time is greater or equal than the TRC's quorum (see [](#quorum)), and a invalid update has been produced and distributed in the network.
 
 ## Initial TRC Signing Ceremony {#trc-ceremony}
 
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -639,7 +639,7 @@ The `version` field describes the version of the TRC format specification. It MU
 
 The `iD` field contains an unique identifier for the TRC, constituted by a sequence of these attributes:
 
-- `iSD`: ISD number,
+- `iSD`: ISD number.
 - `baseNumber`: The base number indicates the starting point of the current TRC update chain. This starting point is the currently valid base TRC, which may differ from the initial TRC in the case of a trust reset.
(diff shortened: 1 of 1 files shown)


oncilla (approved):

nicorusti:
Compiled diff is here:
[Diff_ draft-dekater-scion-pki.txt - draft-dekater-scion-pki.txt.html](https://github.com/user-attachments/files/28593542/Diff_.draft-dekater-scion-pki.txt.-.draft-dekater-scion-pki.txt.html)

nicorusti:
@tzaeschke  I'll merge since Dominik already gave it a look, and I need to start consolidating outstanding PRs.. 
If you see any issues, let me know and we can fix them. In my previous comment you also find the compiled diff, in case it is easier for you too see the changes

nicorusti:
Tilmann Zäschke
It seems this PR 166 has already been merged.

My comments:

I like the updated table 7
I like the new section, it describes when to perform trust reset and why it is needed.
However, I think the new section is missing something:
A description on how a trust reset is performed (or a reference to a section that describes it)
A reference to table 7 as an example of a trust reset
Also, the Changelog references table 6, however in the text table description it is called table 7.

PR #167, opened 2026-06-04
Rename Voting AS to Voter

### LINKED ISSUE #154: Rename Voting ASes (they may not have an AS number)
As reported by @oncilla in #149 

>> 1.1. Terminology
Voting ASes: Those ASes within an ISD that may sign TRC updates. The process of appending a signature to a new TRC is called "casting a vote".

>It is technically possible that voters are not actual routable SCION ASes. That is actually also reflected later on by the recommendation that voters should have an ISD-AS number in their subject, but are not required to have it. If we want to be technically correct, we probably should use a different word than AS. For example, in the SSUN, we have a case where VSE is a voter (and we allocated an AS number), but for the time being they are not routable.

>I am also fine with leaving it as is, just wanted to point this out in case it was not known/obvious.

nicorusti:
Resolves #154 

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.txt&url_2=https://scionassociation.github.io/scion-cppki_I-D/voting/draft-dekater-scion-pki.txt)

### DIFF
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -175,7 +175,7 @@ The SCION trust architecture allows parties that mutually trust each other to fo
 
 To fulfill these requirements, SCION introduces the concept of **Isolation Domains**. An Isolation Domain (ISD) is a building block to support heterogeneous trust while achieving high availability and scalability in its control plane ({{I-D.dekater-scion-controlplane}}). It consists of a logical grouping of SCION ASes that share a uniform trust environment (i.e. a common jurisdiction).
 
-An ISD is governed by one or multiple **Voters**. Furthermore, each ISD has a set of ASes that form the ISD core, known as the **Core ASes**. The set of Core ASes and Voters may be, but do not necessarily have to be the same entities, since Voters do not require an AS number. Governance is implemented by a policy called the **Trust Root Configuration** (TRC), which is negotiated by the Voters and which defines the locally scoped roots of trust used to validate bindings between names and public keys.
+An ISD is governed by one or multiple **Voters**. Furthermore, each ISD has a set of ASes that form the ISD core, known as the **Core ASes**. The set of Core ASes and Voters may be but do not necessarily have to be the same entities, since Voters do not require an AS number. Governance is implemented by a policy called the **Trust Root Configuration** (TRC), which is negotiated by the Voters and which defines the locally scoped roots of trust used to validate bindings between names and public keys.
 
 Authentication in SCION is based on X.509 certificates that bind identifiers to public keys and carry digital signatures that are verified by roots of trust. SCION allows each ISD to define its own set of trust roots, along with the policy governing their use. An ISD's TRC is used for signatures pertaining to information originating from that ISD, such as paths, but for nothing originating outside of the ISD. This ISD-level scoping of trust roots enhances security by strictly limiting effect of a compromise to data originating from the compromised AS.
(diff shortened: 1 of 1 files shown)


nicorusti:
@knmeynell if you are ok with it, feel free to merge it directly

draft-dekater-scion-pki.md
```diff
@@ -175,7 +175,7 @@ The SCION trust architecture allows parties that mutually trust each other to fo
...
 
-An ISD is governed by one or multiple ASes, known as the **Voting ASes**. Furthermore, each ISD has a set of ASes that form the ISD core, known as the **Core ASes**. The set of Core and Voting ASes may be, but do not necessarily have to be the same ASes. Governance is implemented by a policy called the **Trust Root Configuration** (TRC), which is negotiated by the Voting ASes and which defines the locally scoped roots of trust used to validate bindings between names and public keys.
+An ISD is governed by one or multiple **Voters**. Furthermore, each ISD has a set of ASes that form the ISD core, known as the **Core ASes**. The set of Core ASes and Voters may be, but do not necessarily have to be the same entities, since Voters do not require an AS number. Governance is implemented by a policy called the **Trust Root Configuration** (TRC), which is negotiated by the Voters and which defines the locally scoped roots of trust used to validate bindings between names and public keys.
```
knmeynell: comma is not needed here...

_but do not necessarily have to be the same entities since Voters do not require an AS number_

draft-dekater-scion-pki.md
```diff
@@ -1037,7 +1037,7 @@ When an AS joins an ISD, it sends the first CSR out of band to one of the CAs as
 When using this automated in-band renewal process, the request requires two distinct cryptographic signatures to ensure both proof of possession and authorization:
 
 - Proof of possession: the inner PKCS#10 CSR MUST be signed using the newly generated private key corresponding to the requested certificate.
-- Authorization: The AS MUST authenticate the request to the Issuing CA by wrapping the CSR in a CMS SignedData structure (cms_signed_request). This outer CMS structure MUST be signed using the existing private key corresponding to one of the AS's currently active and valid AS certificate.
+- Authorization: The AS MUST authenticate the request to the Issuing CA by wrapping the CSR in a CMS signed-data structure (cms_signed_request). This outer CMS structure MUST be signed using the existing private key corresponding to one of the AS's currently active and valid AS certificate.
```
knmeynell: Don't think a hyphen isn't needed here...

_by wrapping the CSR in a CMS signed data structure_

knmeynell (changes requested):

knmeynell:
FYI, the references to Voting AS have not been changed in the Signing Ceremony appendix.

draft-dekater-scion-pki.md
nicorusti: There are two versions of it in the draft.. I'll double check what is the best and be consistent. Maybe like in -controlplane, we can use: `CMS SignedData`

nicorusti:
> FYI, the references to Voting AS have not been changed in the Signing Ceremony appendix.

Where? I don't see any

nicorusti:
[Diff_ draft-dekater-scion-pki.txt - draft-dekater-scion-pki.txt.html](https://github.com/user-attachments/files/28598307/Diff_.draft-dekater-scion-pki.txt.-.draft-dekater-scion-pki.txt.html)

PR #168, opened 2026-06-04
Dominik & Till review - last points

### LINKED ISSUE #149: Review Dominik

nicorusti:
A few las points.

Resolves #149  

Follow up on trust resets to #166 

[Diff with main](https://author-tools.ietf.org/api/iddiff?url_1=https://scionassociation.github.io/scion-cppki_I-D/draft-dekater-scion-pki.txt&url_2=https://scionassociation.github.io/scion-cppki_I-D/dominik-review/draft-dekater-scion-pki.txt)

### DIFF
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -925,7 +925,7 @@ To verify a TRC update, the relying party MUST perform the following checks:
 - In case of a sensitive update:
     - check that all votes are cast by a sensitive voting certificate.
 
-If one or more of the above checks gives a negative result, the updated TRC SHOULD be rejected.
+If one or more of the above checks gives a negative result, the updated TRC MUST be rejected.
 
 ## Trust Reset {#trust-reset-description}
 
--- a/draft-dekater-scion-pki.md
+++ b/draft-dekater-scion-pki.md
@@ -1323,21 +1323,24 @@ Changes made to drafts since ISE submission. This section is to be removed befor
 {:numbered="false"}
 
 - Draftforge review, sort terminology alphabetically
-- Rename Voting AS to Voter and clarify that it does not require an AS number
+- Review of normative language
+- Rename "Voting AS" to "Voter" and clarify that it does not require an AS number
 - remove trust Hierarchy subsection and redundant code block
 - "Trust Model": reword and shorten section about monopoly/oligopoly
-- "Trust as a function" and "Trust Hierarchy": remove redundant sections, since concepts are also explained elsewhere
+- "Trust as a function" and "Trust Hierarchy": remove redundant sections, since concepts are also explained elsewhere (intro and Ceremony)
 - Certificate validity: align maximum validity recommendations to current practice, clarify margin for AS certificate renewal
 - "Regular Voting Certificate" and "Sensitive Voting Certificate": merge two nearly identical sections into one
-- issuerUniqueID and subjectUniqueID: merge two nearly identical sections into one
-- `authorityKeyIdentifier` Extension: clarify support for `authorityCertIssuer` and `authorityCertSerialNumber` attributes
+- id-at-ia Attribute": reword and clarify that it is optional in voting certificates
+- issuerUniqueID and subjectUniqueID: merge two nearly identical sections into one "Unique Identifiers" section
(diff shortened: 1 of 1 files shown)


nicorusti:
@tzaeschke I don't know what else to write about trust resets, I asked in[ Slack](https://scionproto.slack.com/archives/C8ADA9CEP/p1780589737244109)

tzaeschke (approved):
LGTM

PR #170, opened 2026-06-23
Shrink figure

nicorusti:
As requested by RFC editor


nicorusti:
Screenshot 2026-06-23 at 23 38 03
The ASN.1 module still compiles

